PR #155, opened 2014-12-23
Adding tests for read.phylip and modifying some smaller pieces of code

LinguList:
Although it looks like may files have changed, this is but a small PR with minor modificaitoins. The largest change is that I added tests for read.phylip and modified the phylip-read routine. Further changes include further removement of codecs-import-relics which are now almost all extinguished. I changed util.py in one line, since I want to have only pure lines without linesep when reading in text-files with "lines=true" as kwarg in order to prevent checking for linesep chars in all other routines that deal with text data.


LinguList:
Ok, thought it was ok to merge this. Hope nobody disagrees...

PhyloStar:
I don't disagree.

Taraka Rama.

PR #158, opened 2015-02-23
Tokenizer without orthography profile fails invisibly if regex module isn't installed. 

SimonGreenhill:
Using Wordlist.tokenize() without an orthography profiles files invisibly if regex module isn't installed. Nothing in tokenised, so any output is without tokens. e.g. 

* 1. Make sure regex is not installed:

```
⏚ [simon@minerva5:~/Desktop/lp-err] % python run.py
[WARNING] Module 'regex' could not be loaded. Some methods may not work properly.
[WARNING] Module 'regex' could not be loaded. Some methods may not work properly.
[WARNING] Module 'regex' could not be loaded. Some methods may not work properly.
[WARNING] Module 'regex' could not be loaded. Some methods may not work properly.
```

* 2. Run: 

```
from lingpy import *
wl = Wordlist("DOGON.qlc")
wl.tokenize()
wl.output('qlc', filename="DOGON-tokenised")
```

* 3. Inspect DOGON-tokenised.qlc:

```
...
# DATA
ID      DOCULECT        COUNTERPART     CONCEPT TOKENS
#
1       Toro_Tegu       zɔ́ŋɔ̀    (a) fight, squabble
2       Ben_Tey jáy     (a) fight, squabble
...
```

.. note column 5 'TOKENS' is empty.

* 4. Install regex and rerun the code, and the output looks like this:

```
...
ID      DOCULECT        COUNTERPART     CONCEPT TOKENS
#
1       Toro_Tegu       zɔ́ŋɔ̀    (a) fight, squabble     z ɔ́ ŋ ɔ̀
2       Ben_Tey jáy     (a) fight, squabble     j á y
...
```

The solutions are to either --

1. add `regex` to the requirements for lingpy (easiest, but adds a

2. catch this error. The attached patch does this -- it tries to reimport regex if no orthography_profile is defined, and then raises an ImportError. I also move .grapheme_pattern to be non-defined otherwise, as the original python 're' library won't find anything with the pattern \X.

(I've also changed 'trie' to 'tree' in the inline code comments -- the output refers to this as tree not trie, so I thought consistency was good).

### DIFF
--- a/lingpy/sequence/tokenizer.py
+++ b/lingpy/sequence/tokenizer.py
@@ -41,7 +41,7 @@ class Tokenizer(object):
     -----
 
     The Tokenizer reads in an orthography profile and calls a helper 
-    class to build a trie data structure, which stores the possible Unicode 
+    class to build a tree data structure, which stores the possible Unicode 
     character combinations that are specified in the orthography profile 
     and appear in the data source.
 
@@ -115,7 +115,7 @@ class Tokenizer(object):
 
         # orthography profile processing
         if self.orthography_profile:
-            # read in orthography profile and create a trie structure for tokenization
+            # read in orthography profile and create a tree structure for tokenization
             self.root = createTree(self.orthography_profile)
 
             # store column labels from the orthography profile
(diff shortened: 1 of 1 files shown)


LinguList:
Thanks! Gladly accept the patch.

As to the very issue: You need regex, as it is defined as a possible but not an important dependency, that's why we have the warning. In the longer run, it is planned to make a standalone of the module that is used for tokenizer, since it needs user-defined input for othoprofiles, while lingpy's ipa2tokens function works more flexibel on IPA-strings (orthoprofiles work also on orthography, but you need to write them explicitly). So the solution proposed is in the lines of further development, since we don't need regex as a requirement on the longer run.

So my recommendation here is to tokenize by:

```python
>>> from lingpy import *
>>> wl = Wordlist('input.csv')
>>> wl.add_entries('tokens', 'ipa', lambda x: ipa2tokens(x))
```

Here, you take ipa2tokens and add a new column of "tokens" to the wordlist. If you have ipa-like input data, this normally works with high accuracy.

BTW: this is done automatically if you do not load a wordlist, but lexstat:

```python
>>> lex = LexStat('input.csv')
```

PR #160, opened 2015-03-11
Merging recent stuff containing not necessarily major, but rather minor but anyway quite useful mods in a couple of files

LinguList:
See the commits for details. Most mods are in line with things recently discussed throughout our work on Glottobank and afterwards.


lingpy/sequence/tokenizer.py
```diff
@@ -134,6 +134,7 @@ def __init__(self, orthography_profile=None):
                 self.op_rules = []
                 self.op_replacements = []
                 self._init_rules(self.orthography_profile_rules)
+<<<<<<< Updated upstream
```
xrotwang: What's this?

LinguList: please don't ask :-( I made some bad merge stuff and I am still not really into the whole git stuff as to understand always what is being done. Anyway, I also realized I forgot to commit this file, which is why it failed, since on my system it runs, so hope it'll pass now with the latest commit...

PR #165, opened 2015-08-14
Refactored wordlist and lexstat

xrotwang:
Changes fall in these categories:
- [PEP8](https://www.python.org/dev/peps/pep-0008/) compliance
- condensed code
- usage of functionality from `collections` and `itertools`, e.g. `defaultdict` and `combinations`

In general, I think it is not only verbose but maybe even a performance hit to check `dict` membership by catching `KeyError` on item access.


LinguList:
Great! Thanks a lot!

PR #166, opened 2015-08-16
enhanced test coverage, refactored alignment evaluation classes

xrotwang:
Other than adding tests, the main change is an attempt to refactor the evaluation classes in `lingpy.evaluate` such that
- instance attributes are not defined outside of the class' `__init__` method,
- result caching is implemented using `lingpy.util.cached_property`, rather than checking for attributes.

In general, I think, it must be forbidden to create instance attributes as side effects of method calls, because this makes testing and reasoning about the code almost impossible. Having a method called `get_xxx` which returns `None`, but creates a couple of instance attributes should be considered an anti-pattern.

### DIFF
--- a/lingpy/evaluate/acd.py
+++ b/lingpy/evaluate/acd.py
@@ -1,31 +1,17 @@
 # *-* coding: utf-8 *-*
-# These lines were automatically added by the 3to2-conversion.
-from __future__ import print_function
-from __future__ import division
-from __future__ import unicode_literals
-# author   : Johann-Mattis List
-# email    : mattis.list@gmail.com
-# created  : 2013-03-13 18:31
-# modified : 2013-10-19 10:26
 """
 Evaluation methods for automatic cognate detection.
 """
-
-__author__="Johann-Mattis List"
-__date__="2013-10-19"
-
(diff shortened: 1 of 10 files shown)


LinguList:
Very nice, I'll keep in mind what you mentioned regarding the creating of attributes and classes.

As to cogent: I was forking the tree-class from them and contacted the authors some two years ago, asking them regarding plans to switch to py3, but they didn't have any by then. As a result, I roughly started to adjust the code whenever I found problematic parts, but never did this systematically, also because I didn't understand all of the concepts behind the tree algorithms. So it's very helpful that you started having a look at this part.

Thanks a lot a gain!

xrotwang:
Yeah, I guess there's no way around testing the cogent stuff as well, in particular as long as there is no replacement in sight.

LinguList:
I am currently working on a newick-class for lingpy which is much more
sparse than cogent and should work fine, even more consistently, but
simpler. Since all what LingPy so far uses from cogent's tree class is
just the post-order and pre-order traversals and some subgrouping
functions, it may be possible to replace cogent completely.

The current newick code is here:

https://github.com/lingpy/parsimony/blob/master/newick.py

PR #179, opened 2015-11-02
minor changes, although a lot of scripts involved

LinguList:
the major changes are that I tried to resolve the direct module-loading for parts of the library which are never used. So dictionary and tokenizer are now both no longer imported when typing 
```python
>>> from lingpy import *
```
This is also important, since users who do not want to install regex (not in our official dependencies) always complain about the error-message.

The tests were adapted for this in part, also the test for the lexstat-import function was switched off for python3-5, since it fails there constantly.

One major thing is that I adjusted the algorithm-cython converter. It was not working, in fact, and so one couldn't use it, but nobody seemed to have noticed that. Now, the pyx-file conversion was updated (same as before, basically, but made from scratch), and who runs compilePYX can compile all files to pure c-modules from cython.


xrotwang:
I'd recommend using
```
git pull --rebase ...
```
when synching with upstream. This will prevent the automatic commits of the type
```
Merge remote-tracking branch 'upstream/master'
```

LinguList:
Thanks for the hint, didn't know of this, but it's much more convenient...

PR #189, opened 2016-02-15
Reconfigurations with respect to LingPy-2.4, updates on code, and bugfixes

LinguList:
The major bulk of this is just cleaning (I removed useless directories which nobody every looked at for years). I also re-arranged the "how-to-quote", now putting only collaborators who commited in second half of 2015 to the author board, and all former collaborators as "contributors", since it seems prudent to have specified who is actively working on LingPy. 

I also made first adaptations for handling partial cognates in LingPy, both by allowing for such a format to be read with wordlist and alignments class, but also for having it actively integrated into the align module, which now can handle partial cognates and will only align those parts in the words whcih are cognate, provided, morphemes separation is correctly assigned to the words.


LinguList:
There's a problem on travis with the conda-command. But I have other issues with the tests, so I'll first fix them and then make another PR.

PR #190, opened 2016-02-15
Cleaning up unused folders and providing first directions towards partial cognate handling

LinguList:
This is just before I go too far ahead with my fork:

* clean unused data (mainly the "test" stuff nobody every used for years now)
* change author policy (each new release will only list authors who actively contributed to that release, but former contributors will be mentioned as "contributors")
* add first ideas for handling of partial cognacy, which now works such that Alignments recognises the format and will handle partial cognates by only aligning the parts of the words which are assigned as such, provided the morphemes are corectly assigned.
* some smaller bugfixes which popped up in the issues


LinguList:
BTW: the first fail was mainly due to travis failing to import anaconda, since the folder name changed from "miniconda" to "minoconda2".

Issue #200, opened 2016-03-07
Tests should not clutter the source tree


xrotwang:
Some tests write test output to files in the source tree. These should be removed after testing or preferably be written to temporary directories right away.

PR #201, opened 2016-03-08
Cleaning up tests

### LINKED ISSUE #194: Remove output to stdout when running tests
During tests, all output to `stdout` should be suppressed. Right now, this can partly be achieved by setting the log level in the LingPy config file to `CRITICAL`, but ideally, during test runs local configuration shouldn't interfere. In addition there are quite a few `print` calls left in the code which should be made conditional on log levels as well.

xrotwang:
Closes #194, addresses #200. There is still one test where a file
`lingpy-<date>.qlc` is written. The fact that the exact place is hard
to find highlights the necessity to document side effects like file IO
more thoroughly.

Untracked files:
    lingpy-2016-03-07.qlc

### DIFF
--- a/lingpy/log.py
+++ b/lingpy/log.py
@@ -1,6 +1,7 @@
 """Logging utilities"""
 from __future__ import unicode_literals, print_function, absolute_import, division
 import os
+import sys
 import logging
 from logging.config import fileConfig
 from tempfile import NamedTemporaryFile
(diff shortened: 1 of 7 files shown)


LinguList:
Thanks! I think the untracked thing is in tests/align/test_sca.py, since this is the only thing left that also writes stuff:

https://github.com/lingpy/lingpy/blob/master/lingpy/tests/align/test_sca.py#L107

and as far as I can see, this is not included in your PR.

xrotwang:
But in this line an explicit filename (`test`) is given.

LinguList:
right, sorry for rushing...

xrotwang:
AFAICT it has to be one of the cases where a `dict` is passed as `filename` when creating a `QLCParser` instance.

LinguList:
wow, what a new feeling in running the tests without seeing all these ugly warnings and logs ;-)

LinguList:
okay, found it: it is test_ops.py.

PR #205, opened 2016-03-09
small modifications

LinguList:
basically, only the documentation, and a new way to add matrices initially in plot_heatmap function.


LinguList:
I just realized, that there is still one error.log file written to the tests/ not to tests/output/, but I have no idea why. I ran nosetests from within tests/...

xrotwang:
hm. cannot confirm. This should show up in `git status`, right? AFAICT it would not be hidden by `.gitignore`.

LinguList:
yes, that's how I detected it, since it popped up in "git status"...

LinguList:
okay, false alarm, don't ask me what I was seeing there... Maybe I just
forgot to remove a relic...

PR #210, opened 2016-03-11
misc refactoring

xrotwang:
While I think this PR should be merged, its main intention is to exemplify the various refactorings of LingPy I would apply to make it a codebase that is more modern and easier to maintain.

Using the functionality in the python stdlib's `itertools` and `collections` modules will go along way in that direction, building up and using small but well-motivated helper functions from `lingpy.util` will go even further.

Overall the main goal of these refactorings is cutting down on linecount, which IMHO is way too big for LingPy as a whole and for many modules as well. The worst offender in this respect seems to be the assembling of output, which is intermingled in so many functions and methods. Maybe we should add and use a templating library like Jinja2 or Mako to separate most of this code?

### DIFF
--- a/lingpy/align/_align/confidence.py
+++ b/lingpy/align/_align/confidence.py
@@ -5,7 +5,7 @@ import cgi
 
 from lingpy.sequence.sound_classes import class2tokens, token2class
 from lingpy.settings import rcParams
-from lingpy.util import join
+from lingpy.util import charstring, dotjoin
 
 
 def get_confidence(alms, scorer, ref='lexstatid', gap_weight=1):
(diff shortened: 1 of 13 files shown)


LinguList:
Very cool, thanks a lot!

As to:

> Maybe we should add and use a templating library like Jinja2 or Mako to separate most of this code?

I'm afraid I don't really know what these things do, I know jinjia2, but only for the creation of websites. So you're talking about the handling of file-output in general in LingPy here? Reg. this, I agree having a more modular way of handling things would be useful. I was thinking of having an "output" module in the beginning, but since each class then needed it's output function, and plots and other stuff was added, it started to become difficult to handle, already from the general conception. But I think it's time to try to handle this more consistently.

xrotwang:
I'm talking about a template-based approach to creating formatted output. I.e. rather than have code like in [`basic.ops`](https://github.com/lingpy/lingpy/blob/master/lingpy/basic/ops.py#L253-L404), formatting content, there should be a template, which describes in a more declarative way how the file format looks like. Separating template and data structure which is passed to the template renderer would also make both of these - the format and the data structure - easier to understand.

PR #214, opened 2016-03-14
First attempt to make a workable command line interface for lingpy

LinguList:
The argparse stuff is not fun to write, and it is difficult to handle all the different arguments and possibilities. I'll try to prepare a first howto for the LingPy meeting in Tübingen, so that we have some kind of a demo of what this can already do (which is probably good enough for many cases, regarding the functionality (not necessarily regarding the code and how I implemented it).

anyway, I also modified the setup script, so that it installs automaticall a script at /usr/bin, and this worked out of the box for me, so thanks for the tip, @xrotwang


xrotwang:
I think some of the repetitiveness of this code can be factored out. I'll have a go at it tomorrow. But generally, I guess, user interfaces just are no fun to code - be it cli or GUI :)

What I like is that the cli paradigm forces us to think of the LingPy code as library code, i.e. no `print` in the middle of a method allowed anymore, etc.

LinguList:
I wasn't aware of this command-line possibility via setuptools, and I think this is really great for multiple reasons:

1. showing where the code could still be modularized (e.g., gap opening penalty "gop" as a keyword occurs in every second method, but the defaults may vary in pairwise and multiple, which of course not what one would want), or maybe I'm using the wrong word with "modularizing", but more, that the workflow could be presented in such a way that it is much clearer for the user, but also clearer from the code base
2. it is actually quicker in many cases then opening python and running the stuff, so for quick tests, it will also be very useful as a tool for me personally
3. and the point on print inside the methods you mentioned, since now I can include the print in the UI ;-)

It will be tough, however, to provide similar potential via command line as via the code base, but maybe it is just enough to discuss with the colleagues what they need in a first instance, and not attempt to provide full coverage.

LinguList:
BTW: I was a bit rushing with the PR to have it done before tomorrow. There are some inconsistencies: 
* verbose flag was tested, but only in one function (I like that a lot from other command line tools, but I don't insist on it)
* more args than are actually used are defined in the wordlist() function (I didn't finish all there). Basically, they might as well be deleted, or commented out for the moment. With the last "wordlist" argument, it is nevertheless an open question, how far one would like to go here with the command line (including the plotting of trees, tree calculation, or leaving that for other libaries...).

PR #232, opened 2016-04-16
LexStat pairs: homonyms

pavelsof:
This might be intended behaviour; if so, then consider this pull request a call for better docs :)

The `pairs` property of `LexStat` instances ignores homonyms. Taking the `KSL.qlc` test fixture as an example, the English/German pairs will only include one of the English glosses for _I_ and _eye_, because these have the same IPA transcription. So we will get _I_/_ich_ or _eye_/_Auge_, but not both.

The first commit adds a failing unit test (the test fixture should include 200 pairs instead of 195). The second commit barbarically deletes the problematic code; of course, this code seems to be there for a purpose, but all unit tests are OK, which might be a different problem altogether.

### DIFF
--- a/lingpy/compare/lexstat.py
+++ b/lingpy/compare/lexstat.py
@@ -308,10 +308,7 @@ class LexStat(Wordlist):
                 if i < j:
                     for c in sorted(set(dictA).intersection(dictB)):
                         for idxA, idxB in product(dictA[c], dictB[c]):
-                            dA = self[idxA, "duplicates"]
-                            dB = self[idxB, "duplicates"]
-                            if dA != 1 and dB != 1:
-                                self.pairs[taxonA, taxonB] += [(idxA, idxB)]
+                            self.pairs[taxonA, taxonB] += [(idxA, idxB)]
                 elif i == j:
                     for c in sorted(dictA):
                         for idx in dictA[c]:
--- a/lingpy/tests/compare/test_lexstat.py
+++ b/lingpy/tests/compare/test_lexstat.py
@@ -172,3 +172,12 @@ class TestLexStat(WithTempDir):
         lex.cluster(method='sca', threshold=0.5, ref='cogid')
         self.assertEquals(lex[1, 'cogid'], lex[2, 'cogid'], lex[3, 'cogid'])
         rc(schema='ipa')
+
+    def test_pairs(self):
+        pairs = self.lex.pairs['English', 'German']
+        for key1, key2 in pairs:
+            self.assertEqual(self.lex[key1][0], 'English')
+            self.assertEqual(self.lex[key2][0], 'German')
+            self.assertEqual(self.lex[key1][1], self.lex[key2][1])
+            self.assertEqual(self.lex[key1][2], self.lex[key2][2])
+        self.assertEqual(len(pairs), 200)


LinguList:
This is wanted behaviour. For the simple reason that if you derive a scoring function from all pairs, ignoring homonyms, you will be doomed, since especially full duplicates (as it is also annotated in the code in lexstat) will lead to an overweighing of similarities, and full duplicates are frequent in form of colexifications.

The unittest further might work on the KSL data, but it won't work on other wordlist files, since the index of the language depends upon user-input, but your unittest uses the order in which it is given in the KSL data. The recommended way is to use the inherent index dictionary of all wordlist objects:

```python
>>> wl.header['columnname']
```
which gives you the index of the column name.
As a shortcut, you can likewise use the formula:
```python
>>> wl[key, 'doculect']
```
which is equivalent to:
```python
>>> wl[key][wl.header['doculect']]
```
This is not fully but partially documented on the lingpy.org website, and [this tutorial](http://lingulist.de/documents/tutorials/list-2013-lingpy-datastructures.pdf) may provide further help.

lex.pairs is very specific and only serves to get all unique pairs, that is, the ones which are not duplicated. So the attribute is ONLY there since we need all unique pairs. If you want to calculate all real pairs between two languages in your data, it is recommended to do this actively:

```python
import itertools
pairs = []
dictA = lex.get_dict(doculect='German')
dictB = lex.get_dict(doculect='English')
for c in sorted(set(dictA).intersection(set(dictB)):
    for idxA, idxB in itertools.product(dictA[c], dictB[c]):
        pairs += [(idxA, idxB)]
```
that is, exactly as the change proposed in your PR, without checking for duplicates.

As to your request for better documentation: lex.pairs is part of the LexStat algorithm and using the attribute is part of advanced programming that requires to know parts of the source code in lingpy. As such, it is not possible to make it part of the normal user documentation, but on more explicit tutorials or descriptions of the LexStat algorithm itself. Whether it is possible to realize this fully (and whether it is not just better to read the source code), and whether it is actually something that will be read by more than just one user, these are questions that need to be kept in mind before investing time to document a features that is strictly speaking not what a normal user will or would like to understand.

pavelsof:
Of course, if the `pairs` property is part of the internals, then it should not be documented as part of the public API. Advanced users will end up reading the source code anyway (it is Python after all), so documenting such features is not worth the time investment. Then I can recommend prefixing internal properties and methods with underscore (e.g. `self._pairs` instead of `self.pairs`) in order to make it clear for casual users (such as myself) not to play with what they should not be playing :)

LinguList:
Even better might be a documentation that actually lists the things
which are produced, that is, the attributes which are created, since
they may come in handy. There is a possibility to explicitly explain
attributes like "lex.pairs" in the documentation, but I did not too
often make use of it, since it is tedious.

An example in the lingpy.org documentation is here:

http://lingpy.org/docu/align/lingpy.align.PSA.html#lingpy.align.PSA

There, you see, that attributes can also be readily documented, and
while experienced users may find it practical to have access to these,
beginners will just ignore it.

I'll file this as a documentation issue, to give a list of lexstat's
most important attributes, as we already do in part of some other classes.

Thanks a lot for pointing to this!

PR #234, opened 2016-04-20
Adding documentation and starting to enhance the flexibility of LexStat and other modules with respect to CLDF

### LINKED ISSUE #233: documentation: add attribute-description in lexstat class
Given that we have the possibility to document attributes of a class, this should be also done in the lexstat class, where attributes like:

* lex.pairs

and more should be mentioned in the documentation at http://lingpy.org

LinguList:
This PR contains especially major updates in documentation, although there is still much to be done. It fixes #233 adding documentation on attributes in lexstat. More importantly, the major column names, which were so far hard-coded in LexStat ("tokens", etc.) are now represented as attributes of the lexstat class and can be modified by the user in passing the respective keywords. This is nothing that would be needed to be talked about in the general documentation, but it is important to guarantee flexibility with respect to plugins, but even more important to guarantee flexibility with respect to the CLDF format, where major column names might well differ from lingpy's current practice. It is also a first step towards discarding the whole "alias" system, since the coding of column names as keyword parameters would hopefully allow for the flexibility we need.

As an example:

```python
>>> lex = LexStat("KSL.tsv", segments="clpa_segments")
```
will load the LexStat object and force LexStat to treat the values in column "clpa_segments" as the major segments (what is usually called tokens).

I further refactored (a little bit) the colexification module, which is actually not that bad, but was outdated with respect to very recent changes in python.networkx (they switched to generators for all list-types). As a result, the community-module can no longer be used, since it is not compatible with networkx. Test coverage is here now at 100%. So I hope this will boost the general coverage and make sure that every PR increases coverage.


LinguList:
One further addon: sine MultiTree changed to json-formated trees (compare [http://new.multitree.org//static/data/trees/json/25458.json](http://new.multitree.org//static/data/trees/json/25458.json) as a representation of [http://new.multitree.org/trees/id/25458](http://new.multitree.org/trees/id/25458)), we do not need the tree-functions in convert.py anymore. I further kicked out the consensus tree calculation based on a reference tree, since we lack the tests and it introduces four rather specific but also peculiar functions in convert.tree.py. In the future, this should be replaced by a full-fledged sankoff parsimony, as is already in preparation.

LinguList:
One more thing to note is: http://lingpy.org now redirects to lingpy.github.io, showing the current documentation. Not much has changed, but re-arrangements etc. which are important for an intermediate release of 2.4 are now there.

LinguList:
I merged before I never end up modifying the code. We have coverage of 73% now, I'd suppose, when reaching 75%, we go for version 2.4 along with the refreshed documentation (end of Mai or June may be realistic).

PR #236, opened 2016-04-24
Important changes with more tests and increased consistency in sound classes

LinguList:
This is an important update that for the first time adds more consistent checks of sound class models with increased testing for similar coverage and explicit errors in case of bad data.

It also adds first tests for the cython-package, which were missing so far.


LinguList:
Coverage is almost 80 percent now! We could now say that we release 2.4 as the version that is still backwards-compatible (to the most degree) when having reached 85 in test coverage. All really new things discussed recently will then go in 3.0, but we'll have a stable version also with an updated documentation in the meantime.

LinguList:
For some collaborative reasons, I'll immediately merge this PR, since we need it to be synched for a study.

PR #237, opened 2016-04-27
Partial Cognate Detection Module, Refining Documentation, and Increasing Code Coverage

### LINKED ISSUE #192: LexStat strikes after 269 languages in lex.get_scorer
When analysing more than 260+x languages, there's an overflow error:
```python
>>> OverflowError: integer division result too large for a float
```

First thing to check would be the datatype normally used in get_scorer. I thought, all was based on floats initially, but I'm not entirely sure right now.

LinguList:
This PR changes again a lot of files. Substantial changes are

* added partial cognate detection module in compare/partial with tests
* changed documentation not greatly, but significantly, with respect to version 2.4, apparently now also works with read-the-docs, although their server strikes from time to time, without mock-modules (which is preferred, to allow for full documentation)
* fixes #192, which was due to the progressbar-calculation, but now works fine (tested on  395 languages)

Code-Coverage is now by 85%.


lingpy/compare/partial.py
```diff
@@ -0,0 +1,544 @@
...
+                        self[idxA, self._prostrings][sA[0]:sA[1]],
+                        self[idxB, self._prostrings][sB[0]:sB[1]],
+                    gop,
+                    scale,
+                    factor,
+                    self.rscorer,
+                    mode,
+                    restricted_chars,
+                    1)[2]
+
```
xrotwang: Maybe a better alternative (assigning `lambda`s to a variable is considered bad style):
```python
def function(idxA, idxB, sA, sB):
    if method == 'lexstat':
        args = [...]
    elif method == 'sca':
        args = [...]
    else:
        raise ValueError
    return calign.align_pair(*args)[2]
```

LinguList: will do immediately, thanks!

lingpy/compare/partial.py
```diff
@@ -0,0 +1,544 @@
...
+                # now get the slices with the function
+                slices = _get_slices(tokens, **kw)
+
+                for i,slc in enumerate(slices):
+                    tracer += [(idx, i, slc)]
+                    try:
+                        trace[idx] += [(i, slc, count)]
+                    except KeyError:
+                        trace[idx] = [(i, slc, count)]
+                    count += 1
```
xrotwang: It seems as if `trace` should be a `defaultdict(list)`.

lingpy/compare/partial.py
```diff
@@ -0,0 +1,544 @@
...
+                for i,(idx,pos,slc) in enumerate(trace):
+                    C[idx] += [c[i] + k]
+                if kw['post_processing']:
+                    _g = nx.Graph()
+                    for i,(idx,pos,slc) in enumerate(trace):
+                        _g.add_node((i,idx,pos))
+                    remove_edges = []
+                    for i,n1 in enumerate(_g.nodes()):
+                        for j,n2 in enumerate(_g.nodes()):
+                            if i < j:
```
xrotwang:
```python
for (i, n1), (j, n2) in combinations2(enumerate(_g.nodes())):
   ...
```

lingpy/compare/partial.py
```diff
@@ -0,0 +1,544 @@
...
+            for c in self.rows:
+                idxs = self.get_list(row=c, flat=True)
+                srcs = [self[k, source] for k in idxs]
+
+                # get connected components
+                g = nx.Graph()
+                g.add_nodes_from(idxs)
+                for i, cogsA in zip(idxs, srcs):
+                    for j,cogsB in zip(idxs, srcs):
+                        if i < j:
```
xrotwang:
```python
for (i, cogsA), (j, cogsB) in combinations2(zip(idxs, srcs)):
   ...
```

xrotwang:
Good job! More tests and more docs is perfect!

LinguList:
I realized how encouraging it can be to get those code-pieces covered, especially when the badge from codecov changes the color from deep red to more greenish. 

Documentation is still failing from time to time, btw, but in fact, even when it officially fails, it is well-converted:

* http://lingpylingulist.readthedocs.org/en/latest/

The paper on automatic Partial cognate detection was accepted and needs to be camera-ready on March 19, so I plan to make some 2.4-release which will also be shipped to pypi by then. This is also to calm the many people who have been using the oldest unstable versions pypi. But I'd say, it is also a good point for a stable version before new changes will be introduces which have more impact (like namespaces, etc.).

xrotwang:
May 19, you mean? Make sure to note all necessary steps in a RELEASING.md doc, maybe modelled after [the one for pyramid](https://github.com/Pylons/pyramid/blob/master/RELEASING.txt).

LinguList:
Will do! Probably a bit less verbose (and not with coverage of 100 ;-)),
but having a RELEASE.md is a good idea. I guess, since 2.4.alpha and
2.4.beta, we could as well go for 2.5, to avoid confusion.

PR #249, opened 2016-05-15
Finalizing documentation for 2.5 release

### LINKED ISSUE #84: LexStat: Implement Workflow for Partial Cognate Detection
The workflow could be straightforward:

* identify morphemes in the data or pass it manually
* determine scores between word pairs based on hightest-scoring morpheme scores rather than overall global score between words
* carry out connected-component clustering on the network of connected words in which morpheme matchings are above/beyond the threshold
* align resulting cognate sets, thereby hopefully putting all partial cognate sets into one column or try to determine partial cognate morphemes first and then align all words using a two-step approach

Partial cognate detection is a hot issue and it is not clear whether it can be done in a straightforward way, but it should be followed up since it may help in many, many ways to solve long-standing problems of historical linguistics. When working on this, we also need to take evaluation problems into account (how to judge the quality of partial cognate detection, etc.). All these things could either be implemented as subpart of the LexStat class, or as part of another daughter class of LexStat which allows for more refined modeling techniques.

### LINKED ISSUE #104: Allow for a parsing of alignments and strings containing parts to be ignored
In alignments in historical linguistics, we will often find parts that should be ignored, be it that they represent suffixes, prefixes, or something similar.

For the [EDICTOR](http://github.com/dighl/edictor), a very simple way to represent parts to be ignored in multiple alignments was implemented, where brackets are used to mark parts of sequences as "unalignable":

With this schema, the following alignment

```
SEQA  A B C D
SEQB  - B C -
```

becomes

```
SEQA  ( A ) B C ( D )
SEQB  ( - ) B C ( - )
```

This schema should also be included in LingPy, that is, when parsing already aligned sequences, the internal representation of alignments should exclude these parts.

What is not clear so far, however, is, how to deal with the remainder. Firstofall, we treat it as unaligned, so we can actually just put it at the side. However, when doing alignments with the pre-aligned sequences, it would also be useful to just align alignable parts automatically and then output the whole dump, including the unalignable parts along with the brackets (I am also thinking of unalignable parts marked in the tokens). 

So this needs some more thinking, but unalignable parts should definitely be addressed by LingPy.

### LINKED ISSUE #176: lingpy functionality as UNIX commands
Some of the functionality of LingPy would be worthwhile having available as [UNIX commands](https://en.wikipedia.org/wiki/Pipeline_%28Unix%29), i.e.

- read from `stdin`
- write from `stdout`
- write errors to `stderr`

Relevant functionality:

- segmentation
- cognate detection

### LINKED ISSUE #188: Enable for better specification of file input (wordlist class) errors when there are not enough lines in an input file
Currently, it just throws an index error, but it would be better to point to this explicitly, since most of the time we just know that the file has a different number of columns / cells.

### LINKED ISSUE #244: proposal for print-handling: change all pprint and print-instances with new util.print_stuff function
I added a print_stuff function to util:

https://github.com/LinguList/lingpy/blob/a2b6a121f499e3ed75cc6ee7bffc94375b15bc1e/lingpy/util.py#L179

This function prints stuff unless loglevel prohibits this, and unless one passes the keyword "return_string". Then, it just returns the string.

I applied this to acd.py:

https://github.com/LinguList/lingpy/blob/a2b6a121f499e3ed75cc6ee7bffc94375b15bc1e/lingpy/evaluate/acd.py

and in the test, I receive 100%, since the print-statement is no longer used in the script itself, where we could not cover it, but in utils.

I do not know whether this is the preferred way to do it, but in this way, we could keep the print-functionality, using the keyword "pprint" for interactive sessions, where it is useful, while being able to completely ignore it otherwise. All that functions than need to offer is a pprint-keyword, and a statement that calls utils.print_stuff(string, return_string=return_string).

### LINKED ISSUE #248: line 808 in lexstat should be a keyword
this score for values which are expected according to frequencies but never attested in the data is crucial for the whole process of scorer-calculation and may change results. It should be accessible as a keyword, not set to a constant value of -5, as it is currently done.

LinguList:
I also added some more tests for linkcomm.py in the thirdparty module, which gives us a coverage of about 88% now, but increasing this further will depend from removing the cogent module and replacing it with python-newick as a dependency, as there's no use in writing explicit tests for the lage cogent-phylonode class when we never know whether we'll actually keep it in lingpy.

Documentation is now basically finalized, with very few warnings in sphinx when compiling, and a new section "news" listing major changes, as well as full documentation for the "deep" alignment functions, as it was claimed that it was missing on our earlier workshop in Tübingen.

Many issues are fixed with this PR, or were fixed before, listing them now, to just close them in one rush, instead of having to click through all issues:

* fixes #84, as we have partial cognate detection now
* fixes #188, as now, different line numbers will raise errors
* fixes #104, as this has long since been done
* fixes #176, although we will need to refine this for future versions
* fixes #244, as no more print handlings are left, to my knowledge
* fixes #248, as it is now a keyword

The PR also adds a function (documented and also tested), which allows to read a wordlist using a plain csv file, using the CSV reader (lingpy.basic.wordlist.get_wordlist). A particular problem is, however, that this function needs to use two versions in py3 and py2, since csv-class is handling the stuff really differently in py3 and py2.


LinguList:
Okay, I suggest I close this hear and merge: codecov is at 90%, and I think this is a good start for a release of 2.5, given that documentation is also ready. I suggest I fix the release tomorrow, or a day later this week and make sure to also upload on pypi. If we detect further bugs, we'll have anyway something to build on for the moment, since the current state of the package is worth to be available as something more than just 2.4alphabetagamma, as was the previous release practice...

PR #264, opened 2016-09-30
updating lingpy only slightly, but fixing some annoying errors

LinguList:
some minor fixes


xrotwang:
Guess I need this to run the lexibank cookbooks.

LinguList:
sorry for being reluctant to fixing the coverage... But we'll gain once we remove glosses.py as it is now taken over by pyconcepticon.

xrotwang:
As soon as I get the tests passing again on my system, I'll look into coverage :)
Haven't succeeded yet, though. You went a bit wild with `mock.path`, I guess.

xrotwang:
Also, putting non-ASCII in the message of a `ValueError` is a bad idea for code that should run on py2 as well. Nose will choke on displaying the error, making debugging really hard.

LinguList:
I see. Problem is that those non-ascii things are the things that mess up the correct reading of wordlists, as we have such a huge variation of ways people think they write IPA. But since I generally want to have a more thorough way of checking what is actually wrong in a wordlist when people provide their data, I think we can find a way, maybe a check-command that users can run before even analysing a wordlist, so that the non-ascii all can be removed. It was more a personal convenience, and I didn't really think of potential py2 problems, but I understand of course and see now that it was a very bad idea...

xrotwang:
For these kinds of detailed error reports you should use `log.debug`. For logging, the handling of encodings is far easier to control than for anything printed to stdout.

xrotwang:
Btw.: I can't get `test_sound_class_models` to pass on my system. Any hints?

LinguList:
nope, it's running fine in py2 and py3 on my system. But I had to fix
another bug in my current code. Should I just push what I have until now?

LinguList:
ah, one more point: test_sound_class-models may require you to DELETE
things in your cashe! As the sound class models are all stored there.

xrotwang:
yes, just push.

xrotwang:
ok, `rm .cache/lingpy/*` did the trick. We may have to document this ...

LinguList:
yes, and I'd like to have, but I don't know how to do, a way to clean
the cash, maybe as a command from utils: lingpy.utils.clean_cache(). In
this way, we could say that users should clean the cache after receiving
updates from github?

lingpy/basic/parser.py
```diff
@@ -262,9 +262,17 @@ def __getitem__(self, idx):
...
+        """
+        Modify a specific cell in a specific column of a wordlist.
+        """
+        if not isinstance(idx, tuple) and len(idx) != 2:
+            try:
+                self._data[idx[0]][self.header[self._alias[idx[1]]]] = item
+            except KeyError:
+                    raise KeyError("No line with ID {0} specified could be found.".format(
+                        idx[0]))
+        raise ValueError("__setitem__ requires two values as key.")
```
LinguList: Okay, just installed it. flake8, pycharm will have to wait until Christmas.

PR #274, opened 2016-11-07
Three bug-fixes ...

xrotwang:
... all addressing py2-incompatibility issues.

### DIFF
--- a/lingpy/basic/parser.py
+++ b/lingpy/basic/parser.py
@@ -262,7 +262,8 @@ class QLCParser(object):
                         idx[0])) 
         raise KeyError("No entry with the specified key {0} could be found".format(
             idx))
-    def __setitem__(self, key, item):
+
+    def __setitem__(self, idx, item):
         """
         Modify a specific cell in a specific column of a wordlist.
         """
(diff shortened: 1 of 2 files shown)


lingpy/basic/parser.py
```diff
@@ -364,7 +365,9 @@ def _apply(key, s, *args, **kwargs):
             try:
                 res = function(s, *args, **kwargs)
             except:
-                raise ValueError('Could not convert item "{0}" (ID: {1}).'.format(s, key))
+                print(s, key)
```
SimonGreenhill: stray debug print?

xrotwang: Kind of. This info was previously included in the message passed into `ValueError`; but non-ASCII error messages made nose choke (on py2.7).

LinguList: btw: using the debug functionality would be much better in many cases. What we need, for the users (me included), however, is one short howto that shows how one can switch between the log-levels in an interactive session. I know it's possible, as I have been doing that, but having some easy way to switch, or at least a howto is something I could also offer users who complain that their files don't really load.

PR #275, opened 2016-11-07
make tests pass if optional dependencies are not installed

xrotwang:


LinguList:
good idea, as many of these things should anyway be excluded from lingy's core on the long run. I'm done with investing too much time into visualizations that won't run on other people's computers, or require too much parameter fiddling, especially with matplotlib.

PR #276, opened 2016-11-08
First pass at integrating clldutils

xrotwang:


LinguList:
Is this a clldutils problem, as pathlib is missing as requirement? Anyway, looked at the changes, and I should say: it is such a nice thing to have a well-working path functionality, without all these os.path.XXX commands, this makes all so much easier...

xrotwang:
No. Problem is in one of the tests that require scipy, which don't run on my system. Just in the process of installing scipy.

xrotwang:
pathlib is nice, but just last week I ran into another complication: If you are dealing with non-ASCII path names, they will be `bytes` in py2 and `str` in py3 (on linux - on windows it may be still more complicated). But I still think the only way forward is handling path names preferably in one place only.

LinguList:
completely agree with the paths, I have been profiting from what I learned from you lexibank and concepticon code a lot in setting up the prototypes for clics and the other data-driven projects.

And sorry for Scipy. For quick statistic tests, it is surely nice, but one needs to take in all the monster code, yet for these tests, like ttest or whatever, I don't know how to make a shortcut. But ideally, I'd also like to get rid of numpy in lingpy's core code, as one never knows how stable it will be across platforms.

xrotwang:
Yeah, it's amazing how long scipy takes to compile - and matplotlib to download its 60 MB tarball :)

LinguList:
maybe we could discuss of really out-sourcing all these things into plugins, or also howtos (for the graphics, it's a howto anyway, as one can't have them fit all types of data without fiddling for hours with the parameters). If we put the mpl things into a lingpy/plots package, and some of the difficult things with scipy into a lingpy/calc package, we may even be able to get along without numpy, which will also drastically decrease the loading time of lingpy.

xrotwang:
Hm. I realize that pathlib was just added to the standard lib with py 3.4 - which is the lowest py3 clldutils support. Currently we try to support py 3.3. with lingpy, but I'd say we drop 3.3, since 3.4 seems to be considered the lowest py 3 version to aim at supporting with a single py2/py3 codebase.

SimonGreenhill:
If you're only supporting 3.4+ (good!) then the new [statistics](https://docs.python.org/3/library/statistics.html) package in the stdlib might help cut down the need for some of scipy (not ttest,

xrotwang:
Ok, so merging this would mean bumping the minimum py3 version supported by lingpy to 3.4.

LinguList:
I support that, as I have 3.5 on my linux, and I think, they have also already 3.4 on the mac. And 3.3 is still much slower than 3.4 and 3.5 (judged from gut feeling).

PR #283, opened 2016-11-12
context manager to allow specifying a logging level for a block of code

xrotwang:
Documentation will be added to the logging recipe in the LingPy cookbook.


SimonGreenhill:
cool!

PR #288, opened 2017-01-13
Adding an orthography-profile generator via commandline

LinguList:
This is still a bit experimental, but basically, this creates an initial orthography profile for a given dataset, which is useful for lexibank, etc. Usage with command-line is straightforward:

```shell
lingpy ortho_profile -i input_file --cldf --output_file orthography.prf
```
Do not use --cldf, if input is a standard wordlist (lingpy standard). if you do not specify the output_file, it will default to "orthography.prf".

Some explicit behavior of this profile generation includes:

* leave things with brackets untouched, as in high-quality datasets, the user should decide here what to do
* leave things with "splitters" untouched, that is, slashes, etc. indicating more than one word form, since again, this needs to be decided on a case-to-case basis
* indicate cases of fully retained items as <???>, and of problematic segments as <?>.
* output unicode code-points for all characters.

For a dataset like Galuciotupi, this yields a reasonable amount of 150 segments, with quite some explicit errors. Correcting these cases, however, is very straightforward, and should be doable within less than half an hour. 

See here for a test on this dataset, based on an automatically generated orthography profile, for which unsegmented data was used:

[test-ortho-generation.zip](https://github.com/lingpy/lingpy/files/704897/test-ortho-generation.zip)


xrotwang:
Ah, I thought you were going to implement this as lexibank command.
Am 13.01.2017 18:26 schrieb "Johann-Mattis List" <notifications@github.com>:

LinguList:
Yes, I also thought in the beginning, but then I figured that it is probably more useful from lingpy itself, since I started to copy code from there directly (basically a variation on clean_string function).

LinguList:
The suggestion in our lexibank thread to add stdout and stdin instead of file specifications is a good idea. However, given the current cli, it is not clear how to do that: if there is an input file, this is a path, but stdin needs different treatment in order to be used as a path, and I do not know how to implement that in the current version.

LinguList:
I suggest I'll merge this so we can work with it, and test-coverage is also going higher now.

PR #289, opened 2017-01-16
Tentative model of phonemic distance

tresoldi:
Mattis, I made a matrix and converter file from my tentative model of dimensionless phoneme distance. The model itself, as I stress in my draft on ArXiV, is provisory and a proof-of-concept, and so is this conversion for lingpy (I am not sure if I used a good curve to adapt the values, when running tests for multiple and pairwise alignments it seems to be favoring vowels and giving too much penalty fot gaps). If you think it might be useful, you can merge it; otherwise, feel free do reject the pull request, as it is more of an example of models without classes (each phoneme maps to itself and only to itself).


LinguList:
Thanks! You'll see that the tests fail, which means that the sounds that are handled in the classical sound class models contain elements that your model does not contain or vice verse. The test makes sure that we have the same "expressive power" of handling those sounds in all standard models (dolgo, asjp, sca, and the color model for outputting stuff). You can easily circumvent it by keeping the model as is, but removing it's deeper integration. There is no need to include it in the tests, as it's purpose is not the full coverage as we need for the other models right now (you'll notice that the general strategy for sound class models is to be quite permissive of what can be converted, as people do not adhere to IPA ideals normally). This, I guess should already let the test pass.

I'll leave two notes in your commit, as there is one other concern, but it's better to discuss it over the code. 

In general, I'm quite keen on testing this.

lingpy/tests/data/test_sound_class_models.py
```diff
@@ -7,7 +7,7 @@
 
 class Tests(object):
```
LinguList: Here, I suggest, removing it from tests, as model-tests are designed for the "basic" sound class models (we also do not include "jaeger" here, but have it as a model).

lingpy/tests/data/test_sound_class_models.py
```diff
@@ -7,7 +7,7 @@
...
     # get all segments
-    models = ['sca', 'dolgo', 'art', 'color', 'asjp']
+    models = ['sca', 'dolgo', 'art', 'color', 'asjp', 'tresoldi']
```
LinguList: here again, as mentioned above.

lingpy/data/model.py
```diff
@@ -19,7 +19,7 @@ class Model(object):
     Parameters
```
LinguList: It's nice that you also work on documentation. A mental note for me here is that I should add "jaeger" to this, as this model is also readily available.

LinguList:
In general, if the alignments are less good, this may have to do with the features and the internal things lingpy always does as a default, which may not be what you want (it tweaks the gaps, etc.). The more I think about it, this may actually be a perfect example for our lingpy cookbook, on how to produce a new sound class model and test it for alignments.

LinguList:
You can look [here](https://github.com/lingpy/cookbook), and maybe, before we add this PR (once the tests do not fail anymore), we should start some conversation along with tests over there, and models like yours can also be integrated without having them directly in the library, so we could start from there, test a bit until we find how to switch of those params that overweight the vowels, for example, and then we integrate the model into lingpy proper?

xrotwang:
You may also want to look at how sound class models will be implemented in lingpy3:
- They implement [an interface](https://github.com/lingpy/lingpy3/blob/master/lingpy3/sequence/soundclassmodel.py#L73)
- under which they are [registered](https://github.com/lingpy/lingpy3/blob/master/lingpy3/__init__.py#L14) and can be retrieved explicitly.

LinguList:
Right! given that they are already there in lingpy, with the same format requirements, this could even become an example for lingpy-lingpy3 differences and improvements, etc. I'd say, we leave the cookbook for both lingpy2 and lingpy3, but we indicate in the cookbook what version. A more complete illustration of how to use a new sound class model to align (and what one needs to keep in mind here) may also be useful for discussions of alignment algorithms when porting them to lingpy3.

LinguList:
Okay, just found this line: https://github.com/lingpy/lingpy/blob/master/lingpy/compare/lexstat.py#L342

This means, that sound classes need to be single-symbol items, no clusters allowed. There ARE workarounds, and one could use another model for comparison, etc., but it may be useful for a general discussion to potentially change this when porting to lingpy3:

* sound classes are represented as lists/tuples

The advantage is that a model like the one by Tiago could then be directly adapted in the code, as right now, it does not necessarily work (at least in LexStat, but we have the same in Multiple).

tresoldi:
I had seen the failing test and kept it so that you'd know that there was no full coverage. I was planning to to manually map all the missing phonemes tonight, but I'd prefer removing the model from the test now that I confirmed it is acceptable (as I suspected, due to Jaeger).

As for the alignments, I did some quick investigation of the code and the data (which, by the way, gave me some ideas on how to handle metathesis of phonemes and -- if they will be supported in the future -- of distinctive features). Mielke's matrix, which is my basis, in general gives a larger penalty to vowel->non-vowel transition when compared to other models; this "feature" was reinforced by my arbitrary choices for the missing data-points, as by intuition I know that in-vowel transitions are far more common than out-of-vowel transitions.

I will update the pull request later and start working in its integration into lingpy3 too, if you'd like.

LinguList:
Well, so far, the most striking problem for alignments is that we only allow one symbol on the conversion side, so unless we change this behavior (to be discussed for lingpy3, as it may be useful to use just one representation, such as lists, for all sequences), this model can't be use in lingpy's traditional alignment algorithms, but quite likely by tweaking lingpy, one can write an app for it. I suggest, given that the integration of models is just as easy as:

```python
from lingpy import *
tresoldi = Model('tresoldi') # suppose you have a folder with the model just where you are now
```

We can experiment with this model from within our cookbook, as there I can give answers in detail on how to test the alignment behavior. Later then, we can include it, once we have figured out how to use it for the alignments, and we have lots of test data that can be used in this context. I look forward to giving more explanations with the cookbook, and we'll definitely integrate it later, but let's first figure out how the model works on the alignments, and how we can make sure lingpy's internal parameters don't disturb your main assumptions (there's more than just the scoring function going on right now, as gaps are also specifically calculated, including prosodic context, etc., etc., I'll explain in detail).

tresoldi:
Sounds fine, tell me if I can do something about this. I'll go back to the alignment algorithm to study it, maybe I can start sketching my ideas by extending it.

LinguList:
Maybe, you could write and submit (do you have access?) to the lingpy/cookbook, a first proposal of how to use your model, and the model itself. With some toy alignments. I'll then jump in and show where to tweak, also providing more data. You could just fork the cookbook, open a new folder, put your code in there, and we'll handle it from there?

tresoldi:
Do you mean the repository at https://github.com/lingpy/cookbook ? I can do it, I'd just need to code the tests I did on my own (the Voldemort set and some IE cognates), which is how I found out about the "vowel precedence" and the gap penalty.

By the way, I have removed the model from the test, it is now failing due to a change to clean_string()

LinguList:
exactly. I'll then add more data for you to test, and it will be easier for me to adjust code, and in this way show you where potential pitfalls lie (it is quite likely the vowel-bias is lingpy, not your model).

tresoldi:
I wrote a bare bones notebook, with some fake data for stressing the alignment; by intuition, I'd say that 'sca' is definitely performing better, but on "real" data the results are the same. https://github.com/tresoldi/cookbook/blob/master/multiple-alignments.ipynb

tresoldi:
By playing with the models, it seems the difference is not in gaps or vowels, but in the scores I attributed when comparing any phoneme with the "null" phoneme, so that it performs poorly in deletions and metathesis. There is room for a lot of improvement...

LinguList:
Okay, some initial tests from my side. First, we do it *locally*, so we install the model dynamically, not by adding it to lingpy. Make sure to start python from the same folder where you have the ```tresoldi/``` folder for the model:

```python
from lingpy import *
tresoldi = Model('tresoldi', path='.')
```
Now you have a fully functioning model, but instead of including it in lingpy with a string ```'tresoldi'``` you pass this model directly:

```python
>>> Multiple(strA, strB, model=tresoldi)
```

Now let's test the behavior:

```python
# let's start with a simple test of the scores, as this is important to see what the model does
strA, strB = 'voldemort', 'waldemar'
# let's align the strings in simple algorithm, so we can see how the tresoldi model scores them
almA, almB, score = pw_align(strA, strB)
# now use the tresoldi.scorer to check the scores for the segments
for a, b in zip(almA, almB):
    print('{0}\t{1}\t{2:.2f}'.format(a, b, tresoldi(a, b)))
```
The output is interesting:

```shell
v	w	-4.58
o	a	0.96
l	l	5.00
d	d	5.00
e	e	5.00
m	m	5.00
o	a	0.96
r	r	5.00
t	-	-22.50
```

It shows, that the mapping of "v" and "w" is negatively scored. As a rule of thumb, based on my experience, and this is important for adjusting your model, @tresoldi, you should make sure that consonants get a score of 0 as the lowest. And you should also make sure that vowels vs. consonants are always negative in score. This is a practical decision, based on the segment representation of phonetic strings. Of course, you can say you want to have an alignment that shows transitions between vowels and consonants. In practice, however, you don't want that, since so far, all tests I did disturbed the alignments at some point. Transitions of this kind will need to be handled additionally, e.g., by looking at two or more columns. The same as we do when dealing with metathesis. So as I suggested above: the model performance is some artifact of the lingpy specifics, and the practical need for alignments, not from theory, and more adjustment will be needed.

As a last point, consider representing your sound classes in the converter in only one symbol. E.g., you could just use Chinese character range of unicode, and have this automatically be created (3000 — 303F). That means, converter should convert all two-element symbols to one element:

```text
a : \u3000
aː : \u3001
aˑ : \u3002
aˤ : \u3003
ã : \u3004
b : \u3005
...

You'd need to adjust the matrix as well, but this is important to guarantee that the model works with current lingpy settings. For lingpy3, we should definitely adjust this behavior, but right now, it would be too time-consuming to deal with it. And in the end: who cares about the underlying representation?

LinguList:
One more point: The lowest score you can get as a default is -22.5 in lingpy's scorers (see the score with the gap above, as the "-" is not defined). So the scaling of the numbers needs to be adjusted to that. I'd also suggest to raise the score for identity to 10, as in the SCA model, as this gives more place for finegrained differences.

tresoldi:
Such a detailed feedback, thank you! I will work on this next week.

1. As for dynamically loading the model, I will update the cookbook. I guess the current pull request does not make much sense, perhaps you prefer me to close it?

2. As for the behavior, the scores were in line with what I assumed would be an "average" between the models (I should read your thesis in more detail). I will refit the scores, raising the identity to 10.0, but I am not sure if I will be able to meet the two other requirements, i.e., (a) "the consonants get[ting] a score of 0 as the lowest" (I assume you mean the score for the alignment of two consonants) and (b) the score for a vowel/consonant alignment always negative. I would indeed favor a model that does not penalize vowel/consonant alignments, as you imagined, but here it is not a matter of preference (also because I am aware that this one-way model reflects, at most, probabilities of parallel alignment, and not of phoneme transition, which would require at least a two-way model -- I have a little dream of doing it, building on the work of Jäger and especially Kümmel, but it definitely isn't what I am proposing here). The difficulty resides in the fact that the basis of the data is Mielke's matrix of correspondence, which is aligned with my proposal of having a single model for describing all phones, and, as such, in the fact that the high scores for some vowel/consonants come from his "acoustic" source -- that is the case for lateral approximants and for sonorants in general. I will try to map the scores to a new continuous function (probably a sigmoid with a bias -- what do you suggest as the score for the "most different" pairs, such as /s/ and /m/?) so that both requirements are met as far as possible, but I have already confirmed that I cannot fully satisfy them with this model while keeping it consistent (it might be possible with the ideas for a new model I am slowly working in). Would this be acceptable? I will think about alternatives in the meantime, including a modified version with "manually" corrected scores.

3. As for a single-char symbol for each sound class, this is a no-brainer. I guess I will use the Unicode Private Use Area (U+E000..U+F8FF).

LinguList:
In practice, whether you close the PR or you open a new one is no real difference, also in the cookbook, and we'll merge once it seems okay, but we may close this one for now and keep on discussing in the cookboo.

As to the most dissimilar consonants: I'd give them zero, as consonant-consonant is always better than consonant-vowel.

What you see here, also in the results, is the problem of searching for language-independent, objective criteria for sound change, while they are in fact not necessarily so easy to define. First, sound change may be initiated by speaker or listener. Ohala wrote quite a lot on that: kw sounds similar to p, thus it is a listener-initiated sound change in IE and Greek, while other sound-changes are more based on speakers. Also keep in mind that we're talking about systems in general: listener-based sound change does not require to rely on real acoustics, but more on the perception of a speaker who was raised to speak one particular language. This may diverge from acoustics a lot. For that reason, data-driven or intuition-driven approaches, like the ASJP sound class model or the SCA model have pragmatic advantages. It is an interesting study per se to compare the differences between "objective" measures, and the other ones. It would also be interesting to search for aspects that have most power for historical analysis (voicing difference, is, for example, not important for alignments). A independent model as the one you are trying to build may help to narrow these things down. Maybe it won't improve the alignments, but imagine, you use your model to then infer the rates on a phylogenetic tree of aligned sequences and give an account on which features change most quickly and which are more stable. This would give us some added value, as so far, we simply don't really know why, say, ASJP or SCA work rather well (and the jaeger-model works even better at times, although not tested in lingpy, but by Gerhard Jäger, in his PMI implementation).

So maybe there is no need to adjust the model at this time, but to rather keep the information and see how it can be tested further. 

Anyway, this study will be extremely valuable for our documentation of lingpy, as we can in this way explain users how to create their own sound-class models, and show how they can test they do what they expect, so I'm glad you started working on that, although I'ms sorry I have not enough time to provide more help.

tresoldi:
Just to clarify, as my request might have looked like I was proposing an alternative to SCA or ASJP: this is only a tentative model for phoneme distance, and maybe I should have called it a "proof of concept". The idea was to, while not defending that such universal model exists (and, in fact, I base my score in Mielke's work), offer a language-independent matrix that can be used as a least common denominator when nothing else is available, thus stressing the need for data- and intuition-driven approaches. That is why I had to resort to acoustic signals and perceptions, and I am probably at fault for not discussing properly, neither in the request nor in my draft, that it should be taken as a starting point, a prior, to help identifying user-defined scores that are inconsistent and/or unlikely, but not impossible (as you guessed regarding inferences of rates). I have some other ideas, like "calibrating" the standard models in-class scores with my matrix and calculating the scores from a list of recorded and supposed sound laws (I have been collecting them like stamps), but it really is more a toy model than anything else.

This being said, I will adjust the model to make it as coherent as possible with the ones included in lingpy, it will also help to identify the areas of greatest disagreement in the "phonetic spectrum" among the models. Hopefully, the cookbook with dynamic models will get quite long.

PR #291, opened 2017-02-10
Corrected two typos (missing parentheses).

tresoldi:
There are two missing closing parenthesis in the Workflow Tutorial; in the second one it might be hard for a beginner to understand where the missing one should be.

### DIFF
--- a/doc/source/tutorial/workflow.rst
+++ b/doc/source/tutorial/workflow.rst
@@ -11,7 +11,7 @@ of the :py:class:`~lingpy.tests.util.test_data` function. We use this function t
 
   >>> from lingpy import *
   >>> from lingpy.tests.util import test_data
-  >>> lex = LexStat(test_data('KSL.qlc')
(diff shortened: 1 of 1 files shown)


LinguList:
thanks a lot! This is a bit embarrassing for me, but good that it's corrected now.

LinguList:
it's our fault: we changed the dependency with clldutils, and since then, the clean_string function works differently. Will merge now and later correct this in the test.

LinguList:
found it: split_text in sound_classes also calls strip_brackets, which is of course not the direct behavior we want in the clean_string function.

tresoldi:
I was just going to submit a pull request... Yes, it's due to Robert's changes to split_text() in clldutils.text about a week ago.

LinguList:
just fixed it. We'll update documentation, by the way, with lingpy-2.7, which we'll need to fix soon (ideally in March or so).

LinguList:
Or was it 2.6? I'm getting confused, the weekend is near...

PR #292, opened 2017-02-10
Fixing more typos

tresoldi:
I read all the docstrings and tried the example codes, there were still some typos. I didn't change the references to the cython modules (such as '_calign' for 'calign'), but perhaps these could be changed too.

### DIFF
--- a/lingpy/align/pairwise.py
+++ b/lingpy/align/pairwise.py
@@ -287,7 +287,7 @@ def pw_align(
     --------
     Align two words using the dialign algorithm::
         >>> seqA = 'fat cat'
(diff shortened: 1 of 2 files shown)


LinguList:
Ohoh, does _cython occur in the docu? This should not be shown to the users, as it will only confuse them. Maybe we find a better way to deal with it in lingpy3. I definitely know that there are libraries working in pure python and additional c-code, but my cython-solution is a bit ad-hoc, yet Inever found out to to do it in schoolbook style.

LinguList:
and thanks. This is highly appreciated for the 2.6 version.

PR #293, opened 2017-02-11
More fixes to documentation and data

tresoldi:
Self-explanatory; I still need a second check to methods/classes for cognancy, borrowing, and phylogeny, but haven't found anything.

I also added the datafile used in the Tutorial currently on the site to test_data and corrected an etymology (while Harry is used as a short form for Harold, it is not a cognate).

### DIFF
--- a/doc/source/tutorial/formats.rst
+++ b/doc/source/tutorial/formats.rst
@@ -31,9 +31,9 @@ specifies all entry types in a given dataset::
     11   Woldemort   Владимир      vladimir    Russian      6
     12   Woldemort   Володимир     volodimir   Ukrainian    6
     9    Harry       Harald        haralt      German       7
-    10   Harry       Harry         hæri        English      7
-    11   Harry       Гарри         gari        Russian      7
-    12   Harry       Гаррi         hari        Ukrainian    7
+    10   Harry       Harry         hæri        English      8
+    11   Harry       Гарри         gari        Russian      8
+    12   Harry       Гаррi         hari        Ukrainian    8
 
 This format can be further extended by adding key-value pairs in the lines before the header, such
 as, for example, information regarding the author, the data, or general notes:: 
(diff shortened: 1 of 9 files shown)


doc/source/tutorial/formats.rst
```diff
@@ -31,9 +31,9 @@ specifies all entry types in a given dataset::
     11   Woldemort   Владимир      vladimir    Russian      6
     12   Woldemort   Володимир     volodimir   Ukrainian    6
     9    Harry       Harald        haralt      German       7
-    10   Harry       Harry         hæri        English      7
-    11   Harry       Гарри         gari        Russian      7
-    12   Harry       Гаррi         hari        Ukrainian    7
+    10   Harry       Harry         hæri        English      8
```
LinguList: I'd prefer to leave "harald" cognate with "harry", as etymologically it is the same, right? and it could also be used to later indicate that there is some partial cognacy.

tresoldi: From the dictionaries I checked they are at most what you describe as "oblique" (Harry for Henry from Heimerich; Harold from *waldan), but there is really no reason to be so detailed. I'll revert it later.

There was actually a test case failing due to my changes, so I'm enabling Travis and will start running nose locally before pushing. I'll do it later today, there are more changes to documentation.

lingpy/tests/test_data/harry.msa
```diff
@@ -1,5 +1,5 @@
-file
-test
+Harry Potter Testset
+Woldermort (in different languages)
```
LinguList: change "Woldermort" to "Woldemort" and the checks should all work fine!

doc/source/tutorial/formats.rst
LinguList: I mean: Woldemart is also only obliquely cognate to "Waldemar", right? And the German books have of course also "Woldemort" for translation. But otherwise, we would not have the fun cases to test the alignments, even if they are not used in the documentation (for Harry, I mean). 

BTW: it is extremely difficult to find actual cases from linguistics where you can show all aspects of alignments...

PR #294, opened 2017-02-12
More fixes to documentation/tutorials

tresoldi:
Some more fixes, mostly typos in the documentation.

I reverted my previous changes to the Harry Potter dataset, including the cognancy of Harry/Harold.

I also added a brief explanation to the README.md file on how to install locally in unix systems when there are problems with permissions (last year during the Spring School some people, mostly on Macs, ended up installing with `sudo python setup.py`, which definitely is not the way to go).

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,12 @@ $ cd lingpy
 $ python setup.py install
 ```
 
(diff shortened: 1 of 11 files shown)


LinguList:
nice, thanks a lot!

PR #298, opened 2017-04-18
First work on sanity checks for optimal word list coverage.

LinguList:
These problems are actually not really simple, so I created some simple checks that allow us to test quickly:

* whether minimal mutual coverage in a wordlist is T concepts
* how many languages offer a minimal mutual coverage for T concepts

One can also use the code to compute the average mutual coverage (AMC), which is also a good score, and in general, I suspect the reason why the algos for automatic cognate detection get worse in bigger datasets to be a slow decrease in AMC. Sometimes, it goes down to just some 50 words on average, so no wonder lexstat does not perform well here.

### DIFF
--- a/lingpy/cache.py
+++ b/lingpy/cache.py
@@ -9,9 +9,10 @@ import pickle
 
 from clldutils.path import Path
 from appdirs import user_cache_dir
+from lingpy import __version__
 
 
-DIR = Path(user_cache_dir('lingpy'))
+DIR = Path(user_cache_dir('lingpy', version=__version__))
 
 
 def path(filename):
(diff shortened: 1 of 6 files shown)


LinguList:
I'll merge this now to make it available in other projects. But in general, the code may be enhanced and expanded.

xrotwang:
Yeah, sorry. I'm a bit swamped with work right now. Not much time for reviewing.

LinguList:
no worries. This code is rather straightforward, and I did not mess things up (promise ;-). But I tell you: this was important, as I figured that in some test-datasets the coverage was only 50 words on average, although the start set was some 210. This indicates that mutual coverage is something one should never underestimate, as it will also have an effect on phylo-estimation.

PR #304, opened 2017-09-26
fixing graph-handling due to new networkx release which is backwards-incompatible

LinguList:
Apart from this, this pr provides bug-fixes and a new small module, in which I add random functions for clusters. My main goal was to create a random-linguist-persona which we can use to carry out cognate judgments on data in a random fashion. Probably the best way to compare algorithmic results against: the random linguist, annotating random cognate sets, versus some informed method. This also shows in tough cases, how hard it is to beat the random mode.


xrotwang:
ok, was a bit terrified by 34 changed files, but it turned out to still be reviewable :)

LinguList:
sorry, this also accumulated during the last time: I made minor fixes, but in the end forgot to make PRs in between...

PR #309, opened 2017-10-17
Dead requirements file

SimonGreenhill:
The file [pip-requirements](https://github.com/lingpy/lingpy/blob/master/pip-requirements) seems to be dead? the standard is requirements.txt (which exists) and it has an odd version number in it (numpy>=1.4.0 -- current version is 1.13).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,6 @@
 numpy
 appdirs
-networkx
+networkx>=2.0
+matplotlib>=1.3.0
 clldutils>=1.6
 tqdm
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,3 +4,4 @@ networkx>=2.0
 matplotlib>=1.3.0
 clldutils>=1.6
 tqdm
+six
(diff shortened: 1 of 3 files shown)


SimonGreenhill:
Just noticed that [setup.py](https://github.com/lingpy/lingpy/blob/master/setup.py#L62) also lists requirements. I've rewritten setup.py to read from requirements.txt so there's a single point of reference (=requirements.txt)

requirements.txt
```diff
@@ -1,5 +1,6 @@
 numpy
 appdirs
-networkx
+networkx>=2.0
+matplotlib>=1.3.0
 clldutils>=1.6
-tqdm
+tqdm
```
LinguList: I think that "six" is missing.

SimonGreenhill: Ahh didn't spot that as it's installed by one of the prerequisites too (mock, I think)

SimonGreenhill:
Other possible prerequisites are:

* pyclts (used in sequence/profile.py)
* sklearn
* scipy
* matplotlib
* mpl_toolkit

Should any of these be made default too?

PR #310, opened 2017-10-19
Updated requirements for tutorial.

tresoldi:
Didn't add python-newick (not on repositories) and python-igraph (yet, have to guarantee that compiles). `pycldf` might require Python development libraries installed, is that ok?


LinguList:
thanks! will have a look later on.

PR #311, opened 2017-10-30
adding first nexus export that can be used to support beast

LinguList:
Try the following:

```
>>> from lingpy import *
>>> from lingpy.tests.util import test_data
>>> wl = Wordlist(test_data('GER.tsv'))
>>> from lingpy.convert.strings import write_nexus
>>> write_nexus(wl, ref="cogid", beast=True)
```

So beast=True handles whether concepts will be given an extra line. 

What we'd need is: 

- [ ] check for the characters blog which was now added
- [ ] add a template for nice beast handling

### DIFF
--- a/lingpy/convert/strings.py
+++ b/lingpy/convert/strings.py
@@ -3,6 +3,8 @@ Basic functions for the conversion of Python-internal data into strings.
 """
 from __future__ import unicode_literals
 import re
+from collections import defaultdict
+from clldutils.misc import slug
 
 from lingpy import util
 from lingpy.convert.html import template_path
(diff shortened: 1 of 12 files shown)


lingpy/convert/strings.py
```diff
@@ -403,24 +405,16 @@ def multistate2nex(taxa, matrix, filename='', missing="?"):
 
 
 def write_nexus(
-        taxa, matrix, custom=None, custom_name='lingpy', missing="?", gap="-",
-        template="mrbayes.nex", filename="mrbayes.nex", dtype="RESTRICTION",
-        symbols="10",
-        commands=None, commands_name="mrbayes"):
+        wordlist, ref="cogid", custom=None,
+        custom_name='lingpy', missing="?", gap="-", template="mrbayes.nex",
```
SimonGreenhill: Do we want to hide the template implementation from the user e.g. instead of `template`, have 
`nexus_type = ('mrbayes', 'beast', 'beastwords')`. 

The correct template could be chosen based on that argument?

lingpy/convert/strings.py
```diff
@@ -403,24 +405,16 @@ def multistate2nex(taxa, matrix, filename='', missing="?"):
...
+        filename="mrbayes.nex", dtype="RESTRICTION", symbols="10",
```
SimonGreenhill: Only mrbayes wants dtype=restriction (which is a MrBayes-ism only), while beast wants `standard`. This could be specified in the nexus template directly too.

SimonGreenhill: We should be able to guess at what the symbols are too.

lingpy/convert/strings.py
```diff
@@ -403,24 +405,16 @@ def multistate2nex(taxa, matrix, filename='', missing="?"):
...
+        commands=None, commands_name="mrbayes", beast=False):
```
SimonGreenhill: Ahh, re the `beast` parameter -- another reason to replace that all with `mode`. So that the signature looks like:

```
def write_nexus(
    wordlist, ref="cogid",
    mode='mrbayes',   # either mrbayes, beast or beastwords
    custom=None, custom_name='lingpy',
    missing="?", gap="-", 
    filename="nexus.nex",
    commands=None, commands_name="mrbayes"
):
```

lingpy/convert/strings.py
LinguList: Yes, this seems to be an elegant solution! It would, of course, deprive people who want to flexibly work with this code from the possibility, but on the other hand: those who want flexibility can write their nexus export themselves, the code is not that long to adjust it (and it's almost a standalone).

lingpy/convert/strings.py
LinguList: yes, symbols are only 10 in the current form , given that the matrix in the wordlist is evaluated and created like this. For a multi-state nexus export, we'd need some harder machinery, but this needs to be a separate function anyway.

lingpy/convert/strings.py
LinguList: sounds very reasonable.

PR #312, opened 2017-11-06
write_nexus revisions

SimonGreenhill:
Still some minor tweaks to come.


SimonGreenhill:
Ahh. Crap, tests are failing because unittest on py2.7 doesn't have [AssertRegex](https://travis-ci.org/lingpy/lingpy/jobs/298207952)

LinguList:
```bash
Traceback (most recent call last):

  File "/home/travis/build/lingpy/lingpy/lingpy/tests/convert/test_strings.py", line 204, in test_beastwords

    self.assertRegex(nex, r"1 I_ascertainment,")
```

This happens only in Py2, I didn't check the tests, but it may be easy to fix?

xrotwang:
`assertRegex` was only added in Python 3.1 (see https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRegex).

https://travis-ci.org/lingpy/lingpy/jobs/298207952#L2850

lingpy/convert/strings.py
```diff
@@ -404,40 +401,69 @@ def multistate2nex(taxa, matrix, filename='', missing="?"):
...
+        A string to convert to a nexus safe format.
+
+    Returns
+    -------
+    s : str
+        A string containing a nexus safe label.
+
+    """
+    s = unicodedata.normalize('NFKD', s)
+    s = "".join([c for c in s if not unicodedata.combining(c)])
```
LinguList: If you replace this by:
```python
s = "".join([c for c in s if ord(c)<128 and c not in "();?.\":[]/!"])
```
you can get rid of the next line, plus, Chinese character will deleted. I'd also suggest to finally replace all spaces by an underscore, and to issue a warning if nothing is left, or even better, just raise an exception.

LinguList: Note that exclamation mark was also missing in the list of forbidden things.

LinguList: But in general, I just discussed with @xrotwang to  use clldutils-slug which allows to leave spaces via option already, and simply add another option to leave capitalization, which is probably useful.

LinguList: Addon, clldutils.misc.slug already offers all we need:
```python
>>> slug('Äther Ist däi (bla)', lowercase=False, remove_whitespace=False)
'Ather Ist dai bla'
```

SimonGreenhill: clldutils.misc.slug removes useful punctuation like `_`

LinguList:
I found a workaround here:
* https://stackoverflow.com/questions/40918614/unittest-for-both-python-2-and-3

SimonGreenhill:
Cool, tests now pass.

SimonGreenhill:
...and how's this for `slug`?

lingpy/convert/strings.py
```diff
@@ -404,40 +401,69 @@ def multistate2nex(taxa, matrix, filename='', missing="?"):
     return
 
 
+def nexus_slug(s):
```
LinguList: sorry for bothering on this: but I'd still say: the ideal way is to just use:

```python
slug_languagename = lambda s: clldutils.misc.slug(s, remove_whitespace=False, lowercase=False).replace(' ', '_')
```

LinguList: I mean, not literally, but similarly, lambda of course not needed ;-)

SimonGreenhill: Besser?

SimonGreenhill:
...and this one should make MrBayes work happily now.

PR #314, opened 2017-11-15
Provide a semantically meaningful way of saying 'there are no' (=None) comments.

chrzyki:
(instead of, e.g., entering a random string [or some other technique I haven't noted?]). Additionally, it might also be considered to assume None as default for comment to avoid unintended side-effects.

### DIFF
--- a/lingpy/read/csv.py
+++ b/lingpy/read/csv.py
@@ -67,7 +67,7 @@ def csv2list(
     idx = 0 if header else -1
 
     for i, line in enumerate(infile):
-        if line and not line.startswith(comment) and idx != i:
+        if line and (not line.startswith(comment) if comment else True) and idx != i:
             if strip_lines:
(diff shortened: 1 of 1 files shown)


lingpy/read/csv.py
```diff
@@ -67,7 +67,7 @@ def csv2list(
     idx = 0 if header else -1
 
     for i, line in enumerate(infile):
-        if line and not line.startswith(comment) and idx != i:
+        if line and (not line.startswith(comment) if comment else True) and idx != i:
```
xflr6: how about `if line and (not  comment or not line.startswith(comment)  and ...`?

LinguList:
I guess, this should ideally also be mentioned in the doc-string, thus adding a statement like: "Set the keyword to None if you want to parse all lines of your file."

xflr6:
see #329

PR #315, opened 2017-11-15
close to version 2.6 adding cldf functions, some cleaning and new documentation

LinguList:
many changes are due to documentation, this is always messy, but the new documentation should work, no real errors so far on [lingpy/2.6](http://lingpy.org/2.6/)


LinguList:
I am considering adding some more small tests to increase coverage.  Should we merge first, or would you prefer to have a look, @xrotwang ?

LinguList:
Yes, I agree. I should've made a PR on the code changes last week immediately after I did that. I was also thinking to single out documentation to a DOC folder, as it is not really not of daily concern for us, and I could just gitignore the DOC folder in my lingpy repo.

PR #317, opened 2017-11-16
Added some tests, refactored a few things

SimonGreenhill:
Had some time this morning so I thought I'd poke around in lingpy a bit. 

This PR

1. fixes a name issue (Robison Fould => Robinson-Foulds). Only affects doc strings, no code.
2. Makes the `Tree` __init__ a bit better at guessing if it's been given a filename or a string
3. REMOVES the branch-length distance -- this doesn't seem to be used in the code anywhere, and (more importantly) the implementation is wrong. [Cogent.PhyloNode.distance](http://pycogent.org/cookbook/phylonodes.html?highlight=distance) takes a node as an argument (i.e. so can't compare two trees, but nodes within a tree).

### DIFF
--- a/lingpy/algorithm/_tree.py
+++ b/lingpy/algorithm/_tree.py
@@ -17,7 +17,7 @@ class _TreeDist(object):
     @staticmethod
     def grf(treeA, treeB, distance='grf'):
         """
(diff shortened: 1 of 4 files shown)


LinguList:
Cool, more tests, thanks a lot!

PR #318, opened 2017-11-20
removing documentation

LinguList:
documentation is now here: https://github.com/lingpy/doc

and the direct HTML output is here: https://lingpy.github.io


LinguList:
Don't be scared, it's just removing docu and moving it. As of now, documentation runs smoothly with sphinx-build on my machine. As we have it both on https://lingpy.github.io and on lingpy.org, where I'll just update regularly, I don't think we'll need it for readthedocs anymore. I suppose we had for a publication of lingpy-2.6 by end of the week, potentially earlier. I'll still try to find out whether we can a) close existing issues and b) provide a few more tests to make sure test coverage stays at 90%. If you @xrotwang or you @SimonGreenhill  find burning existing problems, please let me know.

PR #319, opened 2017-11-21
fixe a couple of issues, to make sure we're ready for 2.6

### LINKED ISSUE #261: ZeroDivisionErrro in Alignments
Test to replicate (shoud be added to tests):

``` python
>>> msa = Multiple(['tata', 'AB', 'C'))
>>> msa.prog_align()
```

The cython alignment code needs to be checked for strings with similarity=0 and two unrecognized characters. If this situation is met, it should either abort the alignmetn for this reason explicitly, or abort it in all cases where a string has self-similarity <= 0. One might consider adding a flag to allow for two behaviours, as it may be frustrating if because of one bad stirng the whole alignment strikes. But even better is just a more transparent check when loading the alignments class, which indicates danger if there are too many unrecognized characters.

LinguList:
This is to avoid that it's getting bigger. I hope this is the pre-final one. I suppose I'll need at least one more PR, but only minor things, I promise. This also fixes #261


LinguList:
I'll merge this if it passes, as the coverage is finally a bit higher, close to 90% ;)

Issue #320, opened 2017-11-21
Bug in token2class()?


tresoldi:
I've run into (what might be) a minor bug during lexibank development: if the list of tokens has an entry which is a list of strings with an element of length 0 (like in, snippet from this very case, `[['w', 'i', 'l', 'd', 'oː'], ['r', 'u', 'm', 'u', 'j'], ['']]`), the series of checkings will fail, throwing an uncaught `IndexError` exception. The reason is the faulty entry passing the first condition (not being an empty list) but not the following operation that asks for the first character of the empty string.

The correction would be really simple. One solution would be for the `try` on line 660 ( https://github.com/lingpy/lingpy/blob/master/lingpy/sequence/sound_classes.py#L660 ), besides the `KeyError`, also catch an `IndexError`; another is to catch only the `IndexError` and directly return a `0`, which is what the code will do anyway. This will lead to the intended behaviour, raising a `ValueError: [!] your sequence contains only unknown characters` in `tokens2class()`.

I can do a proper PR if you think this is indeed a bug (it should never happen with well-behaved datasets, and in actually helps in finding problems), maybe waiting for after the next release. The `ValueError` message should also be changed, to "... unknown characters or no characters".

LinguList:
I'd prefer to have this changed, as I wanted this function to either yield "0" or the correct sound class, to make sure we can catch unknown sound classes later. Adding an ```except IndexError: return "0"``` should do, but in lexibank, these entries need to be traced properly, as we'd delete the line completely.

tresoldi:
Good, I'm on it.

PR #321, opened 2017-11-21
Fixing bug in token2class() for empty strings.

tresoldi:
Fixes bug in token2class(), as reported [in issue #320 ](https://github.com/lingpy/lingpy/issues/320)

Please note that the fix actually break a test that was asserting precisely whether IndexError was raised with an empty string (was it once the intended behavior?). I removed this test and added a new one.


LinguList:
I am happy with this behavior: ```token2class``` should not raise an error, as it is the responsibility of ```tokens2class``` to show problems in conversion. And with our new practice that says: if all sequences in the conversion with ```tokens2class``` are "0", raise an error, otherwise not, we can also capture potential alignment errors rather early. Thanks a lot for spotting this!

PR #331, opened 2017-12-15
adding clicks to the lingpy sound class model

LinguList:
I just added click sounds to our sound class models, as they are important as well in the light of CLTS, etc., and they have been ignored for quite some time. Their rendering is still not perfect, but at least most will be linked to a specific class of click sounds which will match with consonants.


lingpy/algorithm/extra.py
```diff
@@ -191,23 +191,18 @@ def infomap_clustering(threshold, matrix, taxa=False, revert=False):
...
+    downloadable from http://igraph.org/.
     """
     if not igraph:
         raise ValueError("The package igraph is needed to run this analysis.")
     if not taxa:
         taxa = list(range(1, len(matrix) + 1))
-
     G = igraph.Graph()
-    vertex_weights = []
+    # add vertices to the graph
```
LinguList: last commit destroyed the adding of vertices to the graph, but they need to be explicitly added in igraph.

LinguList:
I'll merge this, as it does not require additional tests apart from general consistency.

PR #334, opened 2018-01-02
first attempt to fix tests with respect to clldutils >= 2.0.1

LinguList:
This does not yet switch to pytest, and I had to take the MakeTempDir class from the older CLLDUTILS.


LinguList:
wow, did not expect this to be that easy...

Issue #330, opened 2017-12-01
Export nexus for TraitLab


LinguList:
Valentin Thouzeau informed me about necessary conversions:

> It was the file without the concepts, and I had to deleat the words without variability (n° 1 and n° 234) to run BEAST. I read the data with R using the function read.nexus.data from the package ape, then I deleated these words, then I wrote the updated dataset using the function write.nexus.data.

LinguList:
Comment by Guido Grimm on the difficulty of turning one nexus into others:

> Actually, that is very simple these days. Do you know Mesquite? It can read in pretty much any format and spill out the matrix as "simple NEXUS" (which is a clean NEXUS accepted by all NEXUS-compatible programmes including SplitsTree), also have an export for MrBayes, which will add the necessary lines, and "extended" Phylip, the basic format used for many other programmes, extended because the Mesquite PHYLIP export can have more than the original 6-character names. 

It means, we could just add a statement, stick to BEAST and MrBayes export in the current version (one bug in MB export, though) and point user to Mesquite.

SimonGreenhill:
Ugh, Traitlab should handle constant sites -- that should be fixed upstream! But I can easily hook this into the nexus exporter if you want.

LinguList:
Suppose, we do the lingpy-2.6.3 by Tuesday, if it is not too much to add this by then, we could include it in 2.6.3, but otherwise, it can safely wait until later...

SimonGreenhill:
Closing this -- can be reopened if the format isn't correct.

PR #345, opened 2018-03-12
fix for #330

SimonGreenhill:
Please check! I don't have a working copy of traitlab, so I can't test this, but it should be what Traitlab users need.


LinguList:
It looks good, although I can't verify. I suppose we take this, send an example to Robin Ryder, and might adjust later, if things turn out to be missing?

PR #346, opened 2018-03-14
Minor refactor

SimonGreenhill:


LinguList:
This passes the checks, but the `CT_` values are important for the weight calculation, or are they in fact obsolete?

SimonGreenhill:
These are just duplicate dictionary keys, so they should be getting overwritten immediately when the dict is defined, and not visible to the rest of the code, right?

LinguList:
sure, sorry, did only look at the github for the diff... didn't see they are duplicates (how embarassing...)

SimonGreenhill:
I've been playing around with http://codefactor.io which is great for spotting these things!

PR #347, opened 2018-03-14
Small refactor

SimonGreenhill:


lingpy/sequence/sound_classes.py
```diff
@@ -1157,7 +1157,7 @@ def class2tokens(tokens, classes, gap_char='-', local=False):
     """
     if not local:
         out = [t for t in tokens]
-        for i in range(len(classes)):
+        for i in enumerate(classes):
```
LinguList: I'd say, this needs to be:
```python
for i, c in enumerate(classes):
    if c in '-X':
        out.insert(i, gap_char)
```

SimonGreenhill: That looks better.

PR #352, opened 2018-04-09
Refactored algorithm tests, fixed missing tests, added requirements

chrzyki:


chrzyki:
Going to go through the test packages package by package (hope that's not too annoying), since I'm not a big fan of huge refactoring commits (in case something breaks further down the road ...).

Issue #365, opened 2018-04-29
ipa2tokens() fails ungracefully on potentially bad input


tresoldi:
I have been working with some data from Heggarty's [sound comparisons](http://www.soundcomparisons.com/). `ipa2tokens()` is failing in some of the transcriptions, as they have  combining diacritical markers in initial position (an example is the transcription for Nigerian Igbo's English "what", as below):

```python
In [1]: import lingpy

In [2]: test = "͡wɔ̞ˑt"

In [3]: test.encode("unicode_escape")
Out[3]: b'\\u0361w\\u0254\\u031e\\u02d1t'

In [4]: lingpy.ipa2tokens(test)
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-4-7a6bff6aa96d> in <module>()
----> 1 lingpy.ipa2tokens(test)

~/src/lingpy/lingpy/sequence/sound_classes.py in ipa2tokens(istring, **keywords)
    150         # check for combiners next
    151         elif char in kw['combiners']:
--> 152             out[-1] += char
    153             merge = True
    154 

IndexError: list index out of range
```

Both considering this input poorly formed or special enough for having lingpy intentionally failing are good positions -- this test case itself seems wrong, and should be `w\u0361\u0254\u031e\u02d1t`. In such cases, however, it would be better to test for initial "combiners" and explicitly refuse them (or we can just accept them, either as standalone tokens or appending a dummy base glyph -- maybe the empty set? -- as it is not up to `ipa2tokens()` to perform data cleansing).

LinguList:
Agreed. Maybe, the standalone thing is the best, if it's not too difficult to add it to the code (okay, in worst case, one more if-loop plus an additional test...).

tresoldi:
I think I would opt for just having it fail more gracefully. It is hard to conceive a situation where you'd actually need a combiner in the first position, plus it can hardly improve alignment and cognate detection (not to mention that it is hard to spot it without escaping the string),

Still, no matter the solution, we do need one more if-statement in `ipa2tokens()` and one more test. I'll take care of it, we can always move back to non accepting it.

tresoldi:
@chrzyki this is not urgent, should I wait for you to finish implementing the new test framework?

xrotwang:
@tresoldi I guess stuff like this would also be good (edge) test cases for clts, right?

tresoldi:
[...]

PR #366, opened 2018-04-30
Fixes bug in ipa2tokens() when a sequence starts with a combiner.

tresoldi:
This fixes the bug discussed [here](https://github.com/lingpy/lingpy/issues/365). The new behaviour is demonstrated here:

```python
In [1]: import lingpy

In [2]: test = b'\\u0361w\\u0254\\u031e\\u02d1t'.decode('unicode_escape')

In [3]: test
Out[3]: '͡wɔ̞ˑt'

In [4]: lingpy.ipa2tokens(test)
Out[4]: ['∅͡', 'w', 'ɔ̞ˑ', 't']
```

Two things:

- For the time being, it is joining the initial combinator with the null phoneme, as in the example above. It seems must (all?) of us agree that it should fail, throwing a `ValueError` exception with enough information so one can immediately understand the problem (especially valid since it can be difficult to catch an initial combinator by eye); we can easily change this if agreed.
- This is *not* implementing a test for the new `ipa2tokens()` behaviour, as @chrzyki is rewriting the tests. It can either be done later or we can wait so I write a proper test later, as this bug-fix is not urgent. In any case, there is no direct `ipa2tokens()` test in `master` at the moment, the function is only tested as part of `align/test_sca` and `sequence/test_sound_classes`.

### DIFF
--- a/lingpy/sequence/sound_classes.py
+++ b/lingpy/sequence/sound_classes.py
@@ -149,8 +149,18 @@ def ipa2tokens(istring, **keywords):
 
         # check for combiners next
         elif char in kw['combiners']:
-            out[-1] += char
-            merge = True
+            # add the combiner to the previous entry in `out`; if there
+            # is no previous characters (i.e., sequence starts with a 
+            # combiner, which is something we perhaps should not accept, see
+            # discussion on issue #365 on GitHub), append the combiner to
+            # a null phoneme glyph
+            if not out:
+                # empty list, i.e., no previous entry
+                out = ['\u2205' + char]
+                merge = False
+            else:
+                out[-1] += char
+                merge = True
 
         # check for stress
         elif char in kw['stress']:


xrotwang:
Yes, I agree this should rather fail. But the exact semantics my depend on the status of this function with regards to `clts`/`segments`. Should `ipa2tokens` call `segments.tokenize.Tokenizer` with the BIPA profile? If so, what added service should/could it provide? Data cleansing after all?

xrotwang:
Alternatively, `ipa2tokens` could be deprecated, marking it for replacement by external functionality.

LinguList:
I'd say: as long as we don't know how well the bipa-profile performs (remember: ipa2tokens can also handle quite strange input), we should be careful of sending it completely to retirement. But it's open to testing: we have quite a long range of tests for it, right?

xrotwang:
@LinguList I'd consider making the strange cases it handles more transparent (by separating them from the regular stuff handled by BIPA) a first step towards retiring it :)

PR #370, opened 2018-06-20
New ngram module

tresoldi:
This PR adds a module for ngram collection to `lingpy.sequence`, replacing and removing some of the functions in `sound_classes`; tests were also added with 99% of coverage. All the previous functions are still available, such as `bigram()` and `trigram()`, but using this new implementation.

The ngram generation is much faster due to functional programming and Python's lazy evaluation (the core is the `zip(*[list comprehension])` code in the primitive `get_n_ngrams()`). Contexts/ngrams are also always returned as tuples (and not lists, as before), so we can directly hash them for the various stochastic applications, manipulate without need of copying them, and use less memory and less computation (the operations can be compared where `bigram()` was previously used, even though I didn't change the operation of `generate.py` yet). All new functions work with either strings (which are split over spaces) or lists/tuples/iterables with elements of all types except `None` (including non-hasheable ones, i.e., lists of lists).

Three basic types of ngrams are supported:

- "Traditional" ngrams, with left context and optional padding;
- Skip ngrams, as increasingly used in NLP applications, following the description in [Guthrie et al.(2006)](https://homepages.inf.ed.ac.uk/ballison/pdf/lrec_skipgrams.pdf) and the implementation in NLTK; however, by default and unlike these references, I am not allowing more than one gap opening as it makes far more sense for historical linguistics (the standard behavior is available, if desired, and it matches the examples given in Guthrie et al., including the order they are presented -- even though one should not rely in such order)
- Positional ngrams, which are similar to skip ngrams but allow a single gap with variable left and right orders of context; the positional ngrams include the actual transition symbol and its index in the sequence, so we can both build non-memoryless stochastic models (i.e., not following the Markov property) and calculate state-by-state probably when scoring the relative likelihood of a sequence given a set of reference sequences. 

One difference in terms of "traditional" ngram operation is that I am using a single symbol for both left and right boundaries, by default `"$"`. I never found a compelling reason for using two different symbols (it is always obvious if we are dealing with one or the other) and would defend this, but I am of course open to your opinions.


tresoldi:
Pinging @LinguList and @xrotwang , passing tests.

SimonGreenhill:
Cool! Could you use functools.partial to lose all the bi/tri/fourgram code? 

e.g.: 
```python
bigrams = partial(get_n_ngrams, step=2)
```

tresoldi:
@SimonGreenhill I don't see why we couldn't, and it sounds a good idea -- more "elegant" at least, as they say. I only included the bi/tri/four functions because they were already in lingpy.

If anyone agrees, I can change them.

LinguList:
yes, change them, good idea.

LinguList:
However, I'd like to keep the documentation on this, as this goes into lingpy, so if no docstring is supported, I'd shortcut it, if possible, but keep the function structure, as we'd otherwise have to fiddle around with the documentation at lingpy.org too much!

tresoldi:
Partials aren't functions but objects, so we can't have it. But I've kept it in `__doc__` (manually assigning it), I'm almost sure the doc generator will catch it.

Just pushed, passed the tests locally.

lingpy/sequence/ngrams.py
```diff
@@ -0,0 +1,685 @@
+# *-* coding: utf-8 *-*
+"""
+This module provides various methods for generating and
```
LinguList: Note that the documentation requires one single line when introducing a module. So make this a one liner that introduces the module (This moodule provides various methods for generating ngrams) and then add a blank line, and add the more detailed information.

lingpy/sequence/ngrams.py
```diff
@@ -0,0 +1,685 @@
...
+from functools import partial
+
+def _seq_as_tuple(sequence):
+    """
+    Internal function for automatically converting a
+    string sequence to a tuple, if needed.
+
+    Parameters
+    ----------
+    sequence: list or str
```
LinguList: Is this 50 characters per column? Looks very small the text, but our PEP version allows 80. It's easier to read if they are all same size.

lingpy/sequence/ngrams.py
```diff
@@ -0,0 +1,685 @@
...
+# This method with zip, besides returning an iterator as desired,
+# is faster than both the previous lingpy implementation and the
+# one in NLTK; as this is the core of the ngram methods, it is
+# important to have at least this primitive as fast as possible.
+# This is intentionally not defaulting to any value for the
+# order, so that users won't confuse a given order to all
+# orders up to and including the given one.
+def get_n_ngrams(sequence, order, pad_symbol='$'):
+    """
+    Build an iterator for collecting all ngrams of a given
```
LinguList: Same as above: one sentence on one line (use backslash if sentence is longer). Afterwards, the longer explanation.

lingpy/sequence/ngrams.py
```diff
@@ -0,0 +1,685 @@
...
+    #
+    # From which we zip all possible combinations.
+
+    for ngram in zip(*[seq[i:] for i in range(order)]):
+        yield ngram
+
+
+def get_all_ngrams(sequence, orders=None, pad_symbol='$'):
+    """
+    Build an iterator for collecting all ngrams of a sequence
```
LinguList: same here (just saw: my own documentation was also wrong here)

lingpy/sequence/ngrams.py
```diff
@@ -0,0 +1,685 @@
...
+        orders = range(len(seq)+1)
+
+    for order in orders:
+        for ngram in get_n_ngrams(seq, order, pad_symbol):
+            yield ngram
+
+
+def get_skipngrams(sequence, order, max_gaps, pad_symbol='$', single_gap_opening=True):
+    """
+    Build an iterator for collecting all skip ngrams of a sequence
```
LinguList: I'd suggest to write this as `get_skip_ngrams`, to be more consistent.

lingpy/sequence/ngrams.py
```diff
@@ -0,0 +1,685 @@
...
+                              (False,) * gap_width + \
+                              (True,) * (order - left_width) # right width
+
+                    # Iterate over the sequence while applying the pattern;
+                    # `idx` is the starting index in `sequence`, `skip` the
+                    # element index in `pattern`
+                    for idx in range(len_seq-order-gap_width+1):
+                        yield tuple(seq[idx+skip] for skip, keep in enumerate(pattern) if keep)
+
+def get_posngrams(sequence, pre_order=0, post_order=0, pad_symbol='$', elm_symbol='###'):
```
LinguList: `get_pos_ngrams`

lingpy/sequence/ngrams.py
```diff
@@ -0,0 +1,685 @@
...
+    """
+    Build an iterator for collecting all positional ngrams of a sequence
```
LinguList: first part (as mentioned above: one sentence, one line).

lingpy/sequence/ngrams.py
```diff
@@ -0,0 +1,685 @@
...
+    for state_idx, subseq in enumerate(subseqs):
+        yield (
+            # pre-context + element + post_context
+            (subseq[:elem_idx] + (elm_symbol,) + subseq[postctx_idx:]),
+            # transition
+            subseq[elem_idx],
+            # state index
+            state_idx)
+
+def get_all_posngrams(sequence, pre_orders, post_orders, pad_symbol='$', elm_symbol='###'):
```
LinguList: `get_all_pos_ngrams`

lingpy/sequence/ngrams.py
```diff
@@ -0,0 +1,685 @@
...
+    # more functional even in Python, but it would likely complicate the code
+    # too much and unnecessarily.
+    for ngram in chain.from_iterable(ngrams):
+        yield ngram
+
+
+# Define partial functions
+
+bigrams = partial(get_n_ngrams, order=2)
+bigrams.__doc__ = """
```
LinguList: good solution with __doc__, but first sentence (etc., as above)

lingpy/sequence/ngrams.py
tresoldi: I've moved to 80 columns both the documentation and the comments.

lingpy/sequence/ngrams.py
tresoldi: All corrected.

lingpy/sequence/ngrams.py
tresoldi: I think this structure is more consistent, I'm not sure if it is clear what each function does: ("traditional") ngrams, skip ngrams, and positional ngrams are fundamentally different in their usages, so the "grammar" I am using is something like: `get` + (optionally number/size) + type of ngrams. That's why we have:

- `get_n_ngrams` - traditional ngrams of (and only of) order `n`
- `get_all_ngrams` - traditional ngrams of all possible orders
- `get_skipngrams` - skip ngrams of an order `n` and number of gaps `g` (as it is not possible to define skip ngrams without these parameters, `get_ng_skipngrams` seemed redundant, but I can change it)
- `get_posngrams` - positional ngrams of a pre-order `p1` and a post-order `p2` (again, could be changed into `get_p_posngrams`)
- `get_all_posngrams` - positional ngrams of a given set of pre- and post-orders (this is the one which might actually be confusing, as it is not all *possible* positional ngrams, but all positional ngrams *up to* some specified orders)
- bigrams, trigrams, fourgrams - which could, for consistency (but without backwards compatibility) be renamed `get_2_ngrams` or `get_2grams` and so on.

What's our take?

lingpy/sequence/ngrams.py
tresoldi: See above, this would be actually confusing, suggesting that (traditional) ngrams and positional ngrams are the same. But we could go for `get_p_posngrams` or `get_all_posngrams`.

tresoldi:
Pinging @LinguList  for review.

tresoldi:
I have also finished the code for the smoothing methods (pending review and some decision about names and where to put it in lingpy) and have prototyped most of the code for the Ngram modelling (including general and by-state probability scoring, entropy/perplexity, and model serialization; random sequence generation, sequence "correction" [i.e., replacement of most unlikely states by more likely ones, with weighted random sampling] and state prediction by context are still pending).

Maybe I should keep pushing the code to this PR, so it can be reviewed in a single take? There is no rush in merging only this ngram code, and I plan to finish the other parts I've just mentioned by the end of the week.

tresoldi:
Closing the PR until I finish documenting the code and writing tests (sorry for the notifications).

tresoldi:
Pinging @LinguList for initial review.

lingpy/sequence/smoothing.py
```diff
@@ -0,0 +1,760 @@
...
+
+    unobserved_prob: float
+        The log-probability for samples not found in the frequency
+        distribution.
+    """
+
+    # Try to load the required libraries, failing if they are not available.
+    # These are not included in the global imports of this module in order
+    # to reduce the number of dependencies.
+    from scipy import linalg, stats
```
LinguList: sorry, I have to nitpick one more thing, I was not explicit enough, but the way we handle thirdparties that are only essential for a few methods, we do already in teh beginning of the script (here from algorithm/extra.py)

```
try:
    from sklearn import cluster
except ImportError:
    cluster = False
try:
    import igraph
except ImportError:
    igraph = False
```
And then, in the function:

```
    if not igraph:
        raise ValueError("The package igraph is needed to run this analysis.")
    if not taxa:
        taxa = list(range(1, len(matrix) + 1))
```
Earlier, we even logged a warning, using lingpy's logger, but this should only be on debug level, as people were annoyed by getting a warning all the time they loaded the library.

And since numpy is officially still a dependency, you can just import it as you please in the beginning of the script.

tresoldi: Ok for numpy, but I am also using scipy, which is not a dependency.

I will use the solution for third-parties you recommend (and had already recommended, this was a quick fix for later implementing what you had requested, which of course I forgot to do before reopening the PR).

LinguList:
and when this is merged, I'm so looking forward to having a nice intro in our [calc blog](http://calc.hypotheses.org) ;)

tresoldi:
Fixed it. @xrotwang , do you have time to take a general look at what I am proposing? It is a bit different from standard ngram implementations.

lingpy/sequence/ngrams.py
```diff
@@ -0,0 +1,1441 @@
...
+
+    Returns
+    -------
+    out: tuple
+        A tuple of the sequence.
+    """
+
+    # We first check for datatype and then for a space, as the first test is
+    # faster (and evaluation is lazy).
+    if isinstance(sequence, str) and ' ' in sequence:
```
xrotwang: If we still aim at py2 compat (or should we drop it?), this would need to be `six.string_types`, I guess.

lingpy/sequence/ngrams.py
```diff
@@ -0,0 +1,1441 @@
...
+        "Population and weight lengths do not match."
+
+    # Get a random number and see in which bin it falls. We need to use this
+    # logic which is a little more complex than something with randint()
+    # in order to allow for floating-point weights.
+    rnd = [random.uniform(0, cum_weights[-1]) for r in range(k)]
+    less_than = [[cw < r for cw in cum_weights] for r in rnd]
+
+    return [population[lt.index(False)] for lt in less_than]
+
```
xrotwang: `_accumulate` and in particular `_random_choices` look like candidates for a utility module (and may be useful in other places, too).

lingpy/sequence/ngrams.py
```diff
@@ -0,0 +1,1441 @@
...
+            '_smooth_kwargs' : self._smooth_kwargs,
+            '_trained' : self._trained,
+            '_p' : self._p,
+            '_p0' : self._p0,
+            '_l' : self._l,
+            '_l0' : self._l0,
+        }
+
+        with open(filename, 'wb') as handler:
+            pickle.dump(model_data, handler, pickle.HIGHEST_PROTOCOL)
```
xrotwang: I'm a bit on the fence regarding this pickling. It's really hard to support object persistence in this way transparently for the end user. What happens, e.g. when additional non-picklable attributes are added? Then old pickles won't survive. So should a version be added to the pickles? Should there be support to migrate old pickles? Also pickles are not portable between py2 and py3.

I'm leaning towards delegating things like object persistence to the end user - maybe adding a method (or property) here which returns a picklable dict representation of the object (and allowing `__init__` to work with such a representation). The actual pickling/unpickling, i.e. interfering with the files, should be left to the end user. Maybe they have a [ZODB](https://en.wikipedia.org/wiki/Zope_Object_Database) running somewhere and want to use that for persitence ...

xrotwang:
@tresoldi I added a couple of comments - none on the implementation of ngram handling in particular (or as opposed to NLTK), because I don't know too much about this. In general, the code looks pretty clean and well-documented! It's almost a let-down that we still have to write tests for Nose, and not pytest :)

lingpy/sequence/ngrams.py
LinguList: Yes, I'd say: let's stick to py2 compat for the time being. It's no longer difficult to do.

lingpy/sequence/ngrams.py
```diff
@@ -0,0 +1,1441 @@
...
+    except StopIteration:
+        return
+    yield total
+    for element in it:
+        total = func(total, element)
+        yield total
+
+def _random_choices(population, weights=None, cum_weights=None, k=1):
+    """
+    Return a `k` sized list of elements chosen from `population` with
```
LinguList: If this function is in standard from version 3.6, but not in 2.7, we should still keep it, but couldn't we add a 3.6 test that would return the standard function?

One more point: if we put this in a utility module, like suggested by @xrotwang, we'd also document it with sphinx, so again, the first sentence should not fill two lines, only one.

lingpy/sequence/ngrams.py
LinguList: Yes, I agree. We had similar problems with support vector machines: they take ages to be trained, but then, they could only be stored in pickle, which renders them useless, so all users have to run their computer for two weeks in order to be able to use the cognate detection method, although all they create in the end is a matrix.

lingpy/sequence/ngrams.py
tresoldi: Oh yes, I am trying to keep it py2 compatible, only loading `six` is a no-brainer.

lingpy/sequence/ngrams.py
tresoldi: They are really only there due to py2, I am only using them also in py3 because I don't want to check for Python version each time (reminds of C's #ifdef...). I'll move them to `lingpy/utils.py` for the time being.

lingpy/sequence/ngrams.py
tresoldi: I see your point and agree with it. I also hated to use `pickle`, and at first wanted to use at least `json` or `shelve`.

I implemented these methods because they were indirectly requested by the reviewers of a paper, as they were requesting many models trained on many subsets of words (each interaction was taking almost 30 minutes in my laptop). "My" problem is that the end-user here likely wants simple and stupid "save" and "load" methods, and any additional code could be taken as an unacceptable hack (even if the library does *exactly* the same under the hood).

That said, using ZODB is an overkill and no other solution is really good. I will remove these methods and (re)implement them in a class that extends NgramModel (maybe even putting in in third-party later) which we can use in case of submitting code for review. Leaner code in lingpy.

lingpy/sequence/ngrams.py
tresoldi: Agree about keeping: remember that the issue is not only py2 vs. py3, but even old versions of py3 (the Python3 version in the stable long-support Ubuntu, for example, is 3.5 and they rejected proposals for moving forward).

As for using the standard function, I'd prefer not to: code that checks for versions is always ugly in my opinion, and we don't run into any performance issues here (I tested it with some hundreds of thousands random points). This also guarantees that the same seed will produce the same output in the same machine when either py2 or py3 are used.

I've already moved it to utils, I'll fix the docstring.

tresoldi:
And thanks for the comments on my code, @xrotwang ! :smiley:

tresoldi:
Done with the changes, passing tests. Still, no rush to merge it, take all the time you need to review.

tresoldi:
I've decided not to code for the percentile rank function in the main class, as it requires numpy and would better fit the extended class for experiments (like the serialization methods).

On my side, I've finished. @LinguList , I cannot issue a merge (no permission, probably). I'll work on a blog post showing how to use the module (and how it is intended to be different from other libraries/tools, such as NLTK or SRILM) next week.

Issue #369, opened 2018-06-07
Converting from and to CLDF: Any reason not to use pycldf?


Anaphory:
The implementation of `from_cldf` (which looks very much like a class method, all it needs doing is to make `to` its first parameter and put in the right place with the right decorator, by the way) states https://github.com/lingpy/lingpy/blob/a64bc8629006dc31b266bfb4bba6a4bb557b3007/lingpy/basic/wordlist.py#L1096

We could supply a lot of that flexibility by using `pycldf`, is there any reason not to?
My database has some non-standard column names (eg. `Lect_ID` instead of `Language_ID`), so I essentially have code doing something like this already (using LingPy for cognate coding and alignment) and could adapt it quite easily if you want me to. My code would also check for a `cognatesetReference` column in both `FormTable` and `CognatesetTable`.

Anaphory:
I have noticed #367, which would need to be taken into account for any change of `from_cldf`, but I don't fully understand it yet.

LinguList:
> We could supply a lot of that flexibility by using pycldf, is there any reason not to? 
If you have a concrete idea of enhancing, I'd be very pleased!
as to 
> I have noticed #367, which would need to be taken into account for any change of from_cldf, but I don't fully understand it yet.
we have in edictor a very strict format, also in lingpy's partial cognate detection, which, however was decided to be represented differently in cldf and is currently not converted by the function.

Anaphory:
Excellent! I'll have a look.

Anaphory:
I lifted functionality I had used in Beastling before, adapting to the LingPy case.

Check https://github.com/Anaphory/lingpy/commit/daf7a649eaff2e95650f559721a9177dbb001ed7 – tell me about any improvements you deem necessary.

So far I have only tested this by running

    >>> lingpy.Wordlist.from_cldf("lexirumah-data/cldf/Wordlist-metadata.json", col="lect_id", row="concept_id")
    <lingpy.basic.wordlist.Wordlist object at 0x7fde9f26ef98>

and seeing no raised Exception.

Anaphory:
What policy should this use to rename columns? Let's assume

    dataset["CognateTable", "cognatesetReference"].name == "CognateClass"

Then the current code would import this with the name `cognate_cognateclass` and LingPy's default for a column like that is `cogid`, but it might also be partial cognates if I remember the standard correctly.
Another option I thought about was importing this as `cognateclass` given that not other table has a column of this name, so nothing needs to be prefixed.

[...]

PR #375, opened 2018-07-25
Load cldf

Anaphory:
Adresses #369, but not the more detailed tests mentioned there.

### DIFF
--- a/lingpy/basic/ops.py
+++ b/lingpy/basic/ops.py
@@ -449,8 +449,10 @@ def wl2qlc(
                 out += '\t' + text_type(value)
             elif type(value) == float:
                 out += '\t{0:.4f}'.format(value)
+            elif value is None:
+                out += '\t'
             else:
-                out += '\t' + value
+                out += '\t{:}'.format(value)
         out += '\n'
 
     util.write_text_file(
--- a/lingpy/basic/wordlist.py
+++ b/lingpy/basic/wordlist.py
@@ -13,6 +13,9 @@ from clldutils import dsv
 from csvw.metadata import TableGroup
 from unicodedata import normalize
 
+import pycldf
+from pycldf.util import Path
+
 from lingpy.convert.strings import matrix2dst, pap2nex, pap2csv, multistate2nex
 from lingpy.settings import rcParams
 from lingpy.basic.parser import QLCParserWithRowsAndCols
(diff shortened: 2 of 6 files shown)


xrotwang:
Hm. Is there a good reason to merge a PR when CI still fails?

LinguList:
See the issue #376

LinguList:
It was a mistake, but as to #376: @Anaphory, I don't see a reason to actually change the settings. Some people have write access and can push to master, some haven't, they can't. We gave you write access, but we could also revoke it, if you prefer. Protection is not what I'd like.

LinguList:
And I'd even not know how to do it ...

LinguList:
BTW2: @Anaphory, to fix the FileNotFoundError, just grep in lingpy, we have a workaround for that, I'm sure.

Anaphory:
> BTW2: @Anaphory, to fix the FileNotFoundError, just grep in lingpy, we have a workaround for that, I'm sure.

The issue was less that I didn't know how to deal with it, and more that I had forgotten that I was looking for an error I explicitly raised myself, so I had to change two spots. (But yes, using `compat` looks like a more consistent solution.)

LinguList:
> Protecting the branch seems to be straightforward: https://help.github.com/articles/configuring-protected-branches/
But generally, I'd agree to "with great power comes great responsibility". Maybe, though, lingpy is mature enough a project to enforce stricter policies, e.g. "no code enters the master branch without code review" - i.e. only through PR?

I agree with this. Is it easy to set it up?

xrotwang:
So, what I just configured is: The master branch can only be changed by
- merging pull requests
- which pass Travis CI
- and have at least one approving review

Let's see whether this is too strict, and inhibiting fast progress. Ideally, I think, this is what we want. There should be enough lingpy committers now, to make it possible to only merge code that has been looked at by two people.

LinguList:
BTW: @Anaphory, you should be part of the "with contributions by" list in the next release (probably first a bugfix with 2.6.5 or so, later then 2.7). If you introduce more, you can even make it into our editorial board :)

tresoldi:
Sounds good to me, too. For the ngram module we already had this in a way, with Travis and 99% coverage of the new code, a PR, @LinguList formally reviewing/approving and @xrotwang more informally.

xrotwang:
Yes, I think with 7 people having write access, reviewing shouldn't be a bottleneck.

PR #381, opened 2018-08-07
first proposal for type tests

LinguList:
this may even pass py2, but I don't test correctly for unicode etc., yet it proposes a first way to handle the most basic ways in lingpy, comments very welcome.

### DIFF
--- a/lingpy/basictypes.py
+++ b/lingpy/basictypes.py
@@ -37,7 +37,8 @@ class lists(_strings):
 
     def __init__(self, iterable, sep=" + "):
         _strings.__init__(self, str, iterable)
-        self.n = [strings(x) for x in iterable.split(sep)]
+        self.n = [strings(x) for x in (' '.join(iterable) if not
+            isinstance(iterable, str) else iterable.split(sep))]
         self.sep = sep
 
     def __add__(self, other):
--- a/lingpy/basictypes.py
+++ b/lingpy/basictypes.py
@@ -1,0 +1,52 @@
+from array import array
+
+
+class strings(list):
+
+    def __init__(self, iterable):
+        
+        list.__init__(self, iterable if not isinstance(iterable, str) else
+                iterable.split())
+
+    def __str__(self):
+
+        return ' '.join(self)
+
+
+class ints(list):
+
+    def __init__(self, iterable):
+
+        list.__init__(self, [int(x) for x in (iterable if
+           not isinstance(iterable, str) else iterable.split())])
+
+    def __str__(self):
+
+        return ' '.join([str(x) for x in self])
+
+
+class floats(list):
+
+    def __init__(self, iterable):
+
+        list.__init__(self, [float(x) for x in (iterable if
+           not isinstance(iterable, str) else iterable.split())])
+
+    def __str__(self):
+
+        return ' '.join([str(x) for x in self])
+
+
+class lists(strings):
+
+    def __init__(self, iterable, sep=" + "):
+
+        strings.__init__(self, iterable)
+        self.n = [strings(x) for x in iterable.split(sep)]
+        self.sep = sep
+
+    def __str__(self):
+
+        return ' '.join(self)
+        
+        
--- a/lingpy/basictypes.py
+++ b/lingpy/basictypes.py
@@ -16,6 +16,10 @@ class _strings(list):
 
         return _strings(self._type, str(self)+' '+str(_strings(self._type, other)))
 
+    def append(self, other):
+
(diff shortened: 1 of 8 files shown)


lingpy/basictypes.py
```diff
@@ -0,0 +1,52 @@
...
+        
+        list.__init__(self, iterable if not isinstance(iterable, str) else
+                iterable.split())
+
+    def __str__(self):
+
+        return ' '.join(self)
+
+
+class ints(list):
```
xrotwang: should inherit from `array`

lingpy/basictypes.py
```diff
@@ -0,0 +1,52 @@
...
+
+        list.__init__(self, [int(x) for x in (iterable if
+           not isinstance(iterable, str) else iterable.split())])
+
+    def __str__(self):
+
+        return ' '.join([str(x) for x in self])
+
+
+class floats(list):
```
xrotwang: actually, let's try this:
```python
from array import array
from functools import partial

class _array(array):
    def __str__(self):
        return ' '.join([str(x) for x in self])

ints = partial(_array, 'i')
floats = partial(_array, 'f')
```

LinguList: Problem with `array` is here: https://bytes.com/topic/python/answers/484386-inherit-array

I get strange init errors when using this in py3.

LinguList:
```
from array import array
from functools import partial

class _array(array):

    def __init__(self, type_, iterable):
        
        array.__init__(self, type_, iterable)

    def __str__(self):
        return ' '.join([str(x) for x in self])

ints = partial(_array, 'i')
floats = partial(_array, 'f')

```

the error is

```
In [8]: _array('i', [1, 2, 34])
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-8-ca395f82e363> in <module>()
----> 1 _array('i', [1, 2, 34])

/home/jmlist/projects/scripts/LingPy/dev/lingpy/lingpy/basictypes.py in __init__(self, type_, iterable)
      6     def __init__(self, type_, iterable):
      7 
----> 8         array.__init__(self, type_, iterable)
      9 
     10     def __str__(self):

TypeError: object.__init__() takes no parameters

```

xrotwang:
Maybe we should simply go with 
```python
ints = partial(array, 'i')
```
and leave the string representation to a utility function? I'd find that even more transparent:
```python
s = spacejoin(ints([1, 2, 3]))
```

xrotwang:
Ok. But in this case we should inherit from `list` and add **all** relevant `__` methods. I'd hate `+=` behaving differently than `append`.

xrotwang:
So this would then be a `list` with enforced types for members and a custom string representation; I guess using "`<type>`+plural `s`" as naming convention is good enough in terms of transparency.

lingpy/basictypes.py
```diff
@@ -16,6 +16,10 @@ def __add__(self, other):
 
         return _strings(self._type, str(self)+' '+str(_strings(self._type, other)))
 
+    def append(self, other):
+
+        return self + other
```
xrotwang: That's not how `append` is supposed to work, that's `extend`! `append` is `self + [elem]` and returns `None`. I think we should be careful here, to not violate a user's expectation/intuition based on the behaviour of python `list`.

LinguList: yeah, when writing tests, I just learned that the hard way, but this showed me, why writing tests is useful. I also figured I can't do without `super` here, will present my proposal soon.

LinguList: And I decided: if one appends, say, `[1, 2]` to `ints('1 2')` it throws an error, as the type check would fail then.

xrotwang:
To flesh this out, we'd still need `__setitem___`, etc. but that still shouldn't be too heavy weight.

LinguList:
Yes, true. I'd also check for length here, as it should be one segment only. The `lists` type could be more fancy, but I'd make this dependent on how often we encounter this (there are instances, where items are set and a set-item for a complete "morphem" to be introduced would come in handy), in order to be a bit careful.

LinguList:
Short note: in testing also on future code, this turns out to be very useful. I'll close this for the time being. I consider adding one more type, (maybe more, but will keep the range small), for cases of alignments, which we want to model both as "a l ( i g n)" and "a l" , so part in brackets is ignored. This is a small operation, often needed, and if it's packed into an attribute, it's still a simple list, but one can consistently handle it in the Alignments class and beyond.

PR #386, opened 2018-09-21
fix for the partial cognates

LinguList:
note that this only works if you have a clean morpheme segmentation with pluses (preferred).


LinguList:
I'll merge this now, so @Anaphory's issue is solved.

Issue #388, opened 2018-10-30
calign.align_pair can throw 


Anaphory:
When running partial cognate coding with LexStat on ASJP-soundclass data, I get a ZeroDivisionError:
```
2018-10-29 19:14:28,879 [INFO] Calculating random alignmentsfor pair rong1269/rong1269.
Traceback (most recent call last):                                                                                                                      
  File "/vol/home/kaipingga/databases/lexirumah-data/pylexirumah/autocode.py", line 119, in <module>
    args.soundclass, ratio_str))
  File "/vol/home/kaipingga/devel/lingpy/lingpy/compare/lexstat.py", line 286, in __init__
    conf=kw['conf'])
  File "/vol/home/kaipingga/devel/lingpy/lingpy/basic/wordlist.py", line 103, in __init__
    self, filename, row, col, conf or util.data_path('conf', 'wordlist.rc'))
  File "/vol/home/kaipingga/devel/lingpy/lingpy/basic/parser.py", line 432, in __init__
    QLCParser.__init__(self, filename, conf=conf)
  File "/vol/home/kaipingga/devel/lingpy/lingpy/basic/parser.py", line 72, in __init__
    raise IOError("Input file '{0}' does not exist.".format(filename))
OSError: Input file 'lexstats-asjp-5.tsv' does not exist.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3.5/runpy.py", line 184, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.5/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/vol/home/kaipingga/databases/lexirumah-data/pylexirumah/autocode.py", line 124, in <module>
    lex.get_scorer(runs=10000)
  File "/vol/home/kaipingga/devel/lingpy/lingpy/compare/lexstat.py", line 1033, in get_scorer
    self._randist = self._get_randist(**kw)
  File "/vol/home/kaipingga/devel/lingpy/lingpy/compare/lexstat.py", line 861, in _get_randist
    kw['restricted_chars'])
  File "lingpy/algorithm/cython/calign.pyx", line 2772, in lingpy.algorithm.cython.calign.corrdist (lingpy/algorithm/cython/calign.c:28018)
ZeroDivisionError: float division by zero
```

In another run, it happened here:

```
Traceback (most recent call last):                                                                                                                      
[...]

PR #389, opened 2018-10-30
Add velar and uvular implosives

Anaphory:
Extend the standard models by the voiced velar implosive
ɠ and the voiced uvular implosive ʛ.


Anaphory:
Fixes the specific error observed in #388 (not the general error handling)

PR #390, opened 2018-11-05
Fix bug in lingpy.basic.wordlist.from_cldf()

tresoldi:
`.from_cldf()` is currently failing if the list of cognates (by default, `cognates.csv`) exists but is empty.

This is the current situation for some datasets in lexibank and seems to be the intended behavior (i.e., have a `cognates.csv` file even if empty). For example, in WOLD: https://github.com/lexibank/wold/blob/master/cldf/cognates.csv

This simple fix just checks if the result of the dictionary comprehension operating on `tbg.tabledict['cognates.csv']` is not empty.


Anaphory:
That is the module-level `from_cldf()` function, not the `Wordlist` class method, right? Would you mind having a look whether `Wordlist.from_cldf()` behaves better? I wrote that method to support CLDF as intended, not just with default file and column names, so I would hope in can deal with this case, as well.

tresoldi:
Oh, sorry @Anaphory , I missed your comment from last week. I'll check.

LinguList:
Yes, please check, I had lot of fun with Wordlist.from_cldf during the last days, it's the new standard.

LinguList:
I'll merge this now, it seems fine to me, and I have huge trust in Wordlist.from_cldf method, but if the normal from_cldf is okay and with propoer description of the slashing-behavior, it is in my opinion really useful, this PR.

PR #393, opened 2018-11-25
finalizing for version 2.6.4

LinguList:
it is time to make a new release with updated documentation, therefore I'm finalizing this now, also trying to solve deprecation warnings.


LinguList:
Documentation is likewise updated now. See https://github.com/lingpy/doc

LinguList:
I didn't add much for the docu, but you can se it here: https://lingpy.github.io

LinguList:
@tresoldi, please check the modifications on your test. Your test used a unix folder that does not exist on windows, and also not on all unix systems (e.g., not on archlinux), which is why it failed. I also had to rename `get_all_ngrams` for consistency reasons, since my earlier method in `sequence/sound_classes` interprets the question differently by splitting a word into all possible ngrams (useful for morpheme detection), so I call your function now `get_all_ngrams_by_order` which is more consistent with what it does, but a revision of the other function names may also be in order (as well as assert statements in the test, which I deleted for the time being).

LinguList:
@xrotwang, I switched to log.warning as it was mentioned that log.warn will be deprecated in future python versions, and I also changed the few instances of cached_property from clldutils. That's also why the PR is a bit longer, as I had to look for these tiny bits here and there.

tresoldi:
@LinguList I'm checking it right now.

tresoldi:
Did a quick fix of the ngram test, the issue is now mentioned in https://github.com/lingpy/lingpy/issues/394 . Everything else is fine by me.

LinguList:
I'll merge this now, so I can proceed with publishing the documentation. Other wishes will have to wait for version 2.6.5.

PR #400, opened 2019-02-05
Initialization of QLCParser from a list of dictionaries

tresoldi:
This PR allows to initialize an object of class QLCParser (and all classes which inherit it, thus including Wordlist, LexStat, etc.) with a list of dictionaries such as in the example below.

While the "protections" the class has in place for direct data manipulation make sense (you don't want to fiddle with `._data`, `._headers`, etc.), it might be more convenient for debugging and prototyping to manipulate pure Python data, or to just initialize an object from such data. If the PR is accepted, my following plan is to implement a transparent method for casting a QLCParser to an equivalent data structure (a list of dictionaries).

If a column is missing, the casting is just following the standard QLCParser behavior of accepting missing data. The order of the columns is not guaranteed, but given that from version 3.6 Python preserves the order of insertion in a dict (a feature which is now part of the languages as version 3.7), the resulting order of the columns will likely be what the user expects (i.e., it will respect the order found in the data itself as much as possible).

```
        wordlist_data = [
            {'concept': 'I',
            'concepticon_id': '1209',
            'counterpart': 'io',
            'segments': 'i o',
            'doculect': 'Italian',
            'glottocode': 'ital1282',},

            {'concept': 'I',
            'concepticon_id': '1209',
            'counterpart': 'eu',
            'segments': 'e u',
            'doculect': 'Portuguese',
            'glottocode': 'port1283',},

            {'concept': 'THOU',
            'concepticon_id': '1215',
            'counterpart': 'tu',
            'segments': 't u',
            'doculect': 'Italian',
            'glottocode': 'ital1282',},

            {'concept': 'THOU',
            'concepticon_id': '1215',
            'counterpart': 'tu',
            'segments': 't u',
            'doculect': 'Portuguese',
            'glottocode': 'port1283',},
        ]

        QLCParser(wordlist_data)
```


tresoldi:
Travis is failing on Python 3.4 because it cannot install lingpy:

```
Installed /home/travis/miniconda2/envs/condaenv/lib/python3.4/site-packages/UNKNOWN-0.0.0-py3.4.egg
error: The 'configparser>=3.5.0b2' distribution was not found and is required by clldutils
The command "python setup.py install" failed and exited with 1 during .
```

Should I take of that? (In the main repository, of course, I'll sync my PR later)

tresoldi:
It would be the better approach, but I am trying to change the code as little as possible. In fact, the `if` block is building an `input_data` variable to look like it was parser from a dictionary, similar to what you suggest.

Now, how to exactly implement this polymorphism is up to debate. I would prefer to accept this in `__init__` in order to be more transparent (that it, instead of having a class method), but it more of a personal preference and it might lead to people abusing that.

The question is: can we add this "from list of dicts" as I am proposing, implementing it as a class method when all other datatypes are moved (dictionaries, string types, etc.), or should we already move everything?

tresoldi:
@xrotwang as for `configparser`, I can't confirm, but it looks like it was dropped: https://anaconda.org/anaconda/configparser

Can we just move to `configparser>=3.5.0`, i.e., the final non-beta of the same version?

tresoldi:
In any case, I am rerunning travis.

tresoldi:
Same problem with `configparser`, I'll open an issue in `lingpy`.

xrotwang:
@tresoldi yes, can you try with adding `configparser>=3.5.0` in `setup.py`, before the entry for `clldutils`? If this fails, I could also make a bug-fix release of `clldutils` with the updated requirement. Let me know what you need.

tresoldi:
I tried in a container of my own and it seems that `3.5.0` is indeed enough. I've filed an issue: https://github.com/lingpy/lingpy/issues/401 , let me know if I can help.

xrotwang:
Re. "factory method" or "if branch": Probably the if-branch is ok here. But as soon as an `__init__` method does not fit on one screen anymore, I think it's time to refactor. And moving the if-branches to individual factory methods would seem like a natural refactoring to me. Unfortunately it would be a non-backards compatible refactoring.

xrotwang:
Ok, I'll release a bugfix for `clldutils` later today.

LinguList:
I support @xrotwang: make `from_XXX` class functions. 

In a similar way, we could say: initialize a wordlist is in the future only accepting a file, but then we have the `from_dict` function as class list, as well as `from_wordlist`, so we can cast things around?

As to this specific method: I think the naming is not very good, as it is not clear what we are dealing with here. What you basically mean is what `clldutils` returned as a csv.dictionary reader or similar. So you could load that with clldutils (or csvw) and then import into lingpy. But if that's what you want, you should use the correct name for this. @xrotwang, what was the function again in csvw or in clldutils thatt returned exactly that type of ordered_dicts?

tresoldi:
Maybe we allow an object to be empty (no user data), and then implement factory methods *while* keeping the `__init__` check for instance type, at least for the time being? Maybe with `staticmethod`?. I mean, something like:

```
class QLCParser(object):
    def __init__(self, init_data=None):
        if isinstance(init_data, str):
            self.init_from_file(init_data)
        elif isinstance(init_data, list):
            self.init_from_list(init_data)
        else:
            [initialize user data to empty]

    @classmethod
    def init_from_file():
        [code]

    @classmethod
    def init_from_list():
        [code]
```

And so on?

xrotwang:
Regarding going forward: Personally, I find the `__init__` method should always expect Python objects (to make it easier to test, for example). So in this case, `__init__` should accept the `dict`, and all other possible initialization options should be moved to factory methods.

LinguList:
I'm fine with that, @xrotwang. As this will break backwards compat, we may want to start fleshing out the functions and bury them when switching to 2.7? There'd be quite some tutorials to re-write, and documentation as well... But we could already see that `Wordlist.from_file``` is in place.

xrotwang:
@LinguList I think @tresoldi is right in separating input to a method from IO. So saying list-of-dicts is what a certain csv reader returns seems backwards to me. As I said above, having methods accept python objects rather than filenames makes for easier testing, and keeps functionality à la "guess file extension", etc. from creeping in.

tresoldi:
Maybe also splitting into things like `Wordlist.from_qlc()` and `Wordlist.from_cldf()`...

LinguList:
@xrotwang, I did not mean to read from file this list thing, but I meant that there is a terminus technicus being used for that, at least in the code, something like tupledictreader, or whatever, I mean this representation is also the basic one that was used before in concepticon. So it's more the naming of the function than it's functionality.

LinguList:
So far, we have the following use-cases:

1 read wordlist from cldf
2 read wordlist from tsv
3 create wordlist from a dictionary of lists, with header being as a default in the first column as a 0, and keys of the dictionary being identical with the identifiers of the list in tsv format
4 create wordlist from a list of key-value pairs (tuples or dictionaries)
5 create wordlist from an existing wordlist (usually by just accessing the internal `Wordlist._data`, which is identical with No 3, but does not have the 0-header)

Anything more?

tresoldi:
But these are all changes that we do not want for the time being, correct? We should split the decision on this particular PR (which can of course be rejected on many grounds, including theoretical ones), which is intended as a "bonus" that keeps backward compatibility and (hopefully) has no conflicts or unintended behaviors, from the other issues of inheritance/polymorphism that demand a deeper study and discussion about the organization of the code, examples of usage, etc.

LinguList:
Ah, name spaces:

1 `Wordlist.from_cldf`
2 `Wordlist.from_csv` (then, maybe with default to `c="\t"`)
3 `Wordlist(dict)` (with optional specification of the header)
4 no idea
5 `Wordlist.import` or `Wordlist.from_wordlist`

But we could also ignore 5, as the `_data` value keeps the dictionary, so basically, one could just copy from there, or have a `Wordlist.as_dict` function, so calling it would be like `LexStat(wl.as_dict())`.

tresoldi:
@LinguList yes, and as @xrotwang pointed 4 is essentially identical to 3, so that we should try to use the same code (I mean, it can be done with a couple of list comprehensions). 5 is also on the same page.

On thing to discuss is if an empty wordlist is to be allowed (without user data, with zero rows).

LinguList:
No empty wordlist should be allowed.

tresoldi:
Namespace 4 could be `Wordlist.from_list_of_dicts`

LinguList:
the point of a wordlist is: It is a static object. You can add columns, you can manipulate them, but you can't add rows. This was already from the beginning the major feature, as the initialization will calculate the the internal matrix that allows to respresentt the wordlist as columns and rows. If you allow for adding values, you can't control this behavior, so you'd need to re-initialize. Neither deletion nor adding of rows (in the virtual tsv-file) is allowed.

LinguList:
I'd say: make a new PR that allows to go for the backwards incompatible solution, but keeps current behavior, but adds the import functions for the different aspects. Reducing to from_cldf, from_csv, the basic call, and the from_list_of_dicts (but here, as said, I prefer the terminus technicus, if it exists).

tresoldi:
I think the only terminus technicus would be `dataframe`, but I would be against using it -- only in academic terms you can really expect it to be a double linked list (or array) of associative arrays, people will expect either a Pandas or an R object (which we definitely *don't* want).

As for a new PR, you mean closing the current one and having one where the `if isinstance()` blocks in `__init__` call the class methods we were discussing?

LinguList:
and btw: the more I think about this, the more I like the separation of the different functions here. It will make the code much clearer, and we can even go for a smooth transition.

LinguList:
So I propose closing this PR and writing an issue, or four issues, for each of the proposed use cases, which can then be handled again.

xrotwang:
I wouldn't include any code for the incompatible solution yet - rather put a design document somewhere. I think the step towards an incompatible LingPy will be bigger than just "un-comment" some code.

LinguList:
BTW: what I meant was the [`csv.dictReader`](https://docs.python.org/3/library/csv.html#csv.DictReader).

xrotwang:
In this case, I think having both doesn't work, because we are dealing with the `__init__` method, which shouldn't call other methods (of the not-fully initialized object).

tresoldi:
That is what I was suggesting: we move the logic for parsing different kinds of data (now passed in the variable `filename`) to their own class methods, **but** we keep the `if isinstance()` checks in `__init__`, so it keeps backward compatibility.

tresoldi:
Yes, `__init__` better not call other methods, which is why I suggested empty wordlists. The user would need to first create the object and then load the data with whatever class method.

But I know, not the best solution and still break backwards compatibility (unless we keep the `isinstance` there and comment them out in future, uncompatible releases)

LinguList:
`__init__` **should**, and I'll insist on that, take the normal dictionary as we allow it now as well as instance. DictReader output is a special cases which also takes long time to load and is less easy to check.

LinguList:
> I wouldn't include any code for the incompatible solution yet - rather put a design document somewhere. I think the step towards an incompatible LingPy will be bigger than just "un-comment" some code.

But a `from_csv` method would not hurt anybody, right? So would a `from_list_of_dicts`.

tresoldi:
In fact it would not. I will close the PR for the time being and do what you suggest, so we can discuss on actual code later, if that is ok (or just close it yourself).

Issue #402, opened 2019-05-23
segmentation marker should never be scored in cognate detection


LinguList:
When working on SEA languages, I found that the segmentation marker `+` is scored and may yield strange cognate judgments. At least that's my current impression, apart from the fact that cognate judgments may anyway look weird in SEA languages. I have already modified the things locally, but they should be packed in a bugfix-release.

PR #403, opened 2019-05-24
fixing several issues and enhancing cldf parsing

LinguList:
this PR is fixing problems i detected recently (see issue #402), but also updates test, and yields a more consisten handling of name-spaces, and datatypes. Specifically, the `wordlist.rc` is now a separate function, so it is no longer part of the parser, and can thus be used somewhere else.


lingpy/basic/parser.py
```diff
@@ -72,30 +99,33 @@ def __init__(self, filename, conf=''):
...
+        #    self._class[name.lower()] = self._class[name.upper()] = eval(cls)
+        #    self._class_string[name.lower()] = self._class_string[name.upper()] = cls
+
+        #    # add the aliases
+        #    for a in alias.split(','):
+        #        self._alias[a.lower()] = self._alias[a.upper()] = name
+        #        self._class[a.lower()] = self._class[a.upper()] = eval(cls)
+        #        self._class_string[a.lower()] = self._class_string[a.upper()] = cls
+
+        #    self._alias2[name] = sorted(set(alias.split(','))) + [name]
```
xrotwang: Can we remove code rather than comment it? We are using version control, after all.

lingpy/basic/wordlist.py
```diff
@@ -1044,7 +1044,35 @@ def coverage(self, stats='absolute'):
             return sum([a / self.height for a in cov.values()]) / self.width
 
     @classmethod
-    def from_cldf(cls, path, columns=[], filter=lambda row: row["Form"], *args, **kwargs):
+    def from_cldf(
+            cls, 
+            path, 
+            columns=[
```
xrotwang: Isn't a mutable default argument problematic here?

lingpy/basic/wordlist.py
```diff
@@ -1044,7 +1044,35 @@ def coverage(self, stats='absolute'):
...
+                'value',
+                'form',
+                'segments',
+                'language_glottocode',
+                'concept_concepticon_id',
+                'language_latitude',
+                'language_longitude',
+                'cognacy'
+                ], 
+            namespace= {
```
xrotwang: See above: mutable default args are risky.

lingpy/basic/parser.py
LinguList: ohoh, sorry

lingpy/basic/wordlist.py
LinguList: yes, I'd prefer a default, however, as this is getting VERY annoying so far. So I'll put it into a none-phrase.

xrotwang:
Are there tests for all the code in `compare/structure.py`?

lingpy/basic/wordlist.py
xrotwang: here you could simply use a tuple instead of a list.

lingpy/basic/wordlist.py
xrotwang: and here something like a tuple of pairs?

LinguList: Then it's good we cover this anyway, as it was columns = [] before. But now I'll allow for "columns='default'" vs. columns=None.

LinguList:
dammit, I should remove that one, sorry. this was not meant to be added now.

lingpy/basic/wordlist.py
LinguList: can you give a quick example for tuple of pairs?

lingpy/basic/wordlist.py
LinguList: a, I see what you mean now. But is the new solution also fine?

LinguList:
I was switching to 3.5 in tests now. I guess, in general we should start and go for pytest here as well. I just tested with pytest, and it is so much faster, but also more easy to compare, etc. This has to re-run again, as I excluded 2.7 before, but then figured, we do not need to give up py2.7 compability for the time being, but should do so probably at least with verison 3.0 of lingpy.

lingpy/basic/wordlist.py
xrotwang:
```python
namespace=(('key', 'value'), ('key2', 'value2'), ...)
```

xrotwang:
I'm tempted to drop py2 support and switch to pytest for tests right away (for 2.8, I'd say). If this is done, thinking about 3.0 may become somewhat easier.

LinguList:
> I'm tempted to drop py2 support and switch to pytest for tests right away (for 2.8, I'd say). If this is done, thinking about 3.0 may become somewhat easier.

Yes. My plan was to make 2.7 the final version of lingpy, also in part breaking backwards-compat, as I'd by then place the MLNs somewhere else, so they don't bother us too much in lingpy, moving all with @PhyloStar to some new library we call [phylopy](https://github.com/phylopy/phylopy/), so we have tree methods in one place, potentially untested (as it is tradition in the field ;), and use lingpy for the sequence comparison tasks.

xrotwang:
Yes, I think backwards-incompatibility "en bloc", i.e. removing complete subsystems is a lot easier to understand and deal with; so doing this for 2.8 seems ok. It still means this PR would need to work with py2.7 and nosetests :)

LinguList:
The travis for py2 cannot work with networkx, I think, but no idea why.

LinguList:
strangely, py2 does not fail on a fresh virtualenv on my computer.

LinguList:
@xrotwang @tresoldi @chrzyki, if one of you could test the py2 problem of this version, this would be nice, or if you have an idea. There's no need to merge directly, but we should have this passed in about two weeks, as some modifications will be crucial for further maintenance, introduced here, specifically the use of basictypes, to enhance the languagae comparison parts, and the fix of #402.

xrotwang:
Ok, I'll have a look at the py2 behaviour. The latest changes from `master` should be merged, too, right?

xrotwang:
@LinguList I think for the time being - i.e. before lingpy 2.8 and dropping py2 support - the simplest solution would be adding `,<=2.2` to the `networkx` line here https://github.com/lingpy/lingpy/blob/c81742f4b44aafb8489eeab7eac7d6b9ade2bdc4/requirements.txt#L3

PR #406, opened 2019-08-01
standardizing lingpy's behavior for partial cognates and fixing bug in the orhtoprofile creation

LinguList:


LinguList:
build is passing: https://travis-ci.org/lingulist/lingpy/

LinguList:
the checks should not have failed, as they worked on my other account, so I merge this now to have piece, also for the sake of the tutorial.

PR #407, opened 2019-08-08
adding new sound to sound classes and refining name spaces

LinguList:


tresoldi:
I'm ok with the changes, but Travis is strangely failing due to a `Device or resource busy`. Maybe something on their side?

I see no reason for the test to fail, but maybe we should wait at least some hours to see if it's not something on their side.

PR #408, opened 2019-08-08
Fixes bug with cognateset

tresoldi:
Pull request #407 has a bug that prevents using `cogid_cognateset_id` directly, as the column name is missing from the default `columns` parameter.

This small PR fixes the bug and allows to use CLDF datasets directly.

### DIFF
--- a/lingpy/basic/wordlist.py
+++ b/lingpy/basic/wordlist.py
@@ -1059,7 +1059,8 @@ class Wordlist(QLCParserWithRowsAndCols):
                 'concept_concepticon_id',
                 'language_latitude',
                 'language_longitude',
-                'cognacy'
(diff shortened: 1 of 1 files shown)


LinguList:
This was deliberate, as I mentioned in the email: you need to add the columns, as it cannot be assumed on default that the cognatesets will be present in the file, which may raise then errors when trying to store data. I'd even say we should discard the "cognacy" from the default columns, as it is also often not there. 

As we agreed with @xrotwang: for the tutorial, we'll need to do the explicit import, specifying columns from cldf, and this is what one also should do generally, so lingpy imports the minimal, everything going beyond this will require for the time being, to explicitly pass the columns argument.

PR #412, opened 2019-09-19
no dsv

SimonGreenhill:
removed in clldutils 3.0

### DIFF
--- a/lingpy/basic/wordlist.py
+++ b/lingpy/basic/wordlist.py
@@ -9,7 +9,7 @@ import numpy as np
 from collections import defaultdict
 
 from six import text_type as str
-from clldutils import dsv
+from csvw import dsv
 from csvw.metadata import TableGroup
 from unicodedata import normalize
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,8 +5,9 @@ nose
 tqdm
 six
 mock
-pycldf
 csvw<=1.5.6; python_version < "3.5"
 csvw>=1.5.6; python_version >= "3.5"
 clldutils==2.8.0; python_version < "3.5"
 clldutils>=2.8.0; python_version >= "3.5"
+pycldf==1.7.0; python_version < "3.5"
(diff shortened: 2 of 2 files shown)


LinguList:
do we have csvw as a dependency yet? It was a dep of clldutils, but is this save?

LinguList:
A more important question: @xrotwang, clldutils now dropped support for py2, right? Does this mean we now need to drop support in lingpy as well? If so, given that the py2 test fails, due to missing pathlib, we can't go on with a version 2.6.5, as planned, but should go to 2.7 right away, I guess?

xrotwang:
Before version 3, clldutils provided a wrapper dsv module, importing csvw.
In lingpy py2 compat can be kept, if clldutils 2.8 is required for py2 < 3

Johann-Mattis List <notifications@github.com> schrieb am Do., 19. Sep.
2019, 14:05:

xrotwang:
@lingulist see pyglottolog 's setup.py for how to make clldutils still work
with py2

Robert Forkel <xrotwang@googlemail.com> schrieb am Do., 19. Sep. 2019,
14:26:

xrotwang:
I'm sorry for this hassle. Specifying a particular major version of
clldutils using the "~=" syntax would have prevented this hiccup. But otoh
it would have prevented upgrade of clldutils for the many cases where the
few breaking changes were not an issue.
So I don't know what the perfect solution is. Probably releasing a bugfix
for the latest minor versions and otherwise urge people to upgrade
everything- I.e  the R solution.
Regarding reproducibility of older studies it probably means that they
should always contain the complete output of pip freeze ...

Robert Forkel <xrotwang@googlemail.com> schrieb am Do., 19. Sep. 2019,
14:29:

LinguList:
Okay, is it possible, @SimonGreenhill, to adjust this in your PR? Since as of now, we loose py2 compat...

SimonGreenhill:
fixed!

PR #413, opened 2019-10-11
bugfix for release of 2.6.5

LinguList:


LinguList:
I'll upload this to pypi then directly, once we have gone through this. The documentation is already online for version 2.5.

LinguList:
Yes, you're right, it is a bit difficult to keep up, when I keep finding things that don't work, but I guess, this version is stable enough by now, and the documentation is ready, so I can upload it. One major addon is, as you [can see here](http://lingpy.github.io/news.html), the new `get_partial_scorer`, which is in fact quite important, in my opinion. 

I have quite a few new things I want to add for the next release, such as:

* parsimony implementation
* a conversion from partial to normal cognates (new algorithm)
* some major tools shipped currently with lingrex, like template-alignment (short scripts, so no difficult code)

I'll make the release now, and I hope that the new version, be it 2.7 or 2.6.6, can be done in less than a year this time...

SimonGreenhill:
Looks good to me!

PR #429, opened 2020-11-25
Allows un-pinning networkx

### LINKED ISSUE #417: Allow networkx > 2.2
Curently, `networkx` is pinned to `<=2.2`, but version 2.4 of that library is already out. AFAICT that's mostly due to using `Graph.node[..]` to access nodes, which can be replaced with `Graph[...]` (at least for newer versions).

xrotwang:
Let's hope our test coverage is high enough to make such a change ...

closes #417

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -32,4 +32,4 @@ jobs:
     - name: "Upload coverage to Codecov"
       uses: "codecov/codecov-action@v1"
       with:
-        fail_ci_if_error: true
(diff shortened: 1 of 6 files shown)


LinguList (approved):
I think they are ;)

xrotwang:
Hope the travis-requirement for PRs goes away soon. I've unregistered the repo at travis, removed `.travis.yml`, and deleted the webhook ...

xrotwang:
Ah, it was hidden in "settings -> branches". Should be good now.

PR #434, opened 2021-06-04
Don't use deprecated stuff from clldutils

### LINKED ISSUE #433: path_component deprecated in clldutls 3.9
With PY3 it's no longer necessary to cast path components to `str`, thus we don't need `clldutils.path.path_component` anymore
https://github.com/lingpy/lingpy/blob/2929d0493aa19e011bb9e4e1741f75977aa84fc1/src/lingpy/util.py#L13

xrotwang:
closes #433


xrotwang:
That's a bit of a catch 22: If you change the ci targets in a PR, the PR will never get a "all checks complete" :)

xrotwang:
Hm. py3.5 isn't in the gh actions yaml in master, either. Wonder why the result is expected.

LinguList (approved):
Thanks! Looks fine to me. I can prepare a release later this weekend and then let you know once it is done (do you have access to the lingpy on pypi, @xrotwang?)

xrotwang:
Yes, I'm listed as maintainer on PyPI

LinguList:
So feel free to also twine this without me, but otherwise, I'd do it 
tomorrow (on the road back to Europe now).

xrotwang:
merged now - can you do the release?

xrotwang:
@LinguList what about a release? Or should we wait til end of June in Leipzig?

xrotwang:
Ah, already done! Sorry.

LinguList:
Yep. There was at first some misunderstanding from my side: as you have also access, I thought, I'd just let you do it whenever you want, and then I realized that I had more time that day than expected and just released it. But I apparently did not think of letting you know.

PR #442, opened 2023-04-30
Fixes #423 (Do not re-segment in ipa2tokens when being passed a list-like object)

### LINKED ISSUE #423: Do not re-segment in `ipa2tokens` when being passed a list-like object
Currently, the code re-segments even when being given a list. This leads to strange results, especially since geminate sounds will be merged. 

https://github.com/lingpy/lingpy/blob/4d42cda2ecd7b54abaf1a2d4aa517c54ff89b573/lingpy/sequence/sound_classes.py#L112-L120

So this line should also have a check for a list or basictypes as input and return them before segmenting what has been segmented already.

PatrickElmer:
This PR fixes issue #423 by checking if the input `istring` variable is an instance of `list` or `tuple` and returning `istring` as a `list` if that is the case.

It also fixes a typo in the `setup.py` that prevented the tests from running and updated the docstring (default values for `merge_vowels` and `merge_geminates` were incorrect).

### DIFF
--- a/src/lingpy/evaluate/alr.py
+++ b/src/lingpy/evaluate/alr.py
@@ -84,8 +84,11 @@ def mean_edit_distance(
         log.debug('{0}, {1}'.format(proto, consensus))
 
         if tokens or classes:
-            proto = ipa2tokens(proto, **keywords)
-            consensus = ipa2tokens(consensus, **keywords)
+            # Only tokenize proto and consensus if not yet tokenized
+            if not isinstance(proto, list):
+                proto = ipa2tokens(proto, **keywords)
+            if not isinstance(consensus, list):
+                consensus = ipa2tokens(consensus, **keywords)
 
             if classes:
                 proto = tokens2class(proto, keywords['model'], stress=keywords['stress'],
--- a/src/lingpy/evaluate/alr.py
+++ b/src/lingpy/evaluate/alr.py
@@ -13,8 +13,6 @@ def mean_edit_distance(
         gold="proto",
         test="consensus",
         ref="cogid",
-        tokens=True,
-        classes=False,
         **keywords):
     """
     Function computes the edit distance between gold standard and test set.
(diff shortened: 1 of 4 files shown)


LinguList (changes requested):
Thanks, the change in setup.py is of course important and the documentation lists the wrong default values. As to the other aspect, I am not sure if we want this behavior. This would only extend the permissiveness of the ipa2tokens function, namely that it allows to pass lists to it, while we could well argue that we should rather only allow a string as input here, given that the function works with regular expressions and the like. So I'd be inclined to say: let us add type hints and raise a value error if one passes a list, instead of returning a list, if the function is being passed a list. With these modifications (maybe the type hint is enough, but I am not against throwing an error upon input that is not a string), I'd merge this.

PatrickElmer:
Thank you for your feedback. I agree that it would be best to raise a value error to prevent anything other than strings from being passed into the function.

When I tried to implement this, I got a test failure in `tests/evaluate/test_alr.py::test_med`. (As far as I can tell, the `mean_edit_distance` function passes a list of tokens into `ipa2tokens` if `classes=True` is set.) If this is intentional, it might be best to allow passing lists and returning them unchanged at the start of the `ipa2tokens` function?

LinguList:
That would be something we need to fix then.

PatrickElmer:
`ipa2tokens` is now raising a value error when not passed a string.
This PR also fixes `mean_edit_distance` passing lists to `ipa2tokens`.

PatrickElmer:
`ipa2tokens` is now raising a value error when not passed a string.
This PR also fixes `mean_edit_distance` passing lists to `ipa2tokens`.

src/lingpy/evaluate/alr.py
```diff
@@ -84,8 +84,11 @@ def mean_edit_distance(
         log.debug('{0}, {1}'.format(proto, consensus))
 
         if tokens or classes:
```
LinguList, line 86: There's an easier solution: we drop the whole line with "if tokens or classes". The function is not used in any publication, I doubt anybody will use it according to new developments, where we model proto-forms on an extra line, like from another language, so we can later even drop the method as a whole and later add more refined ones (https://aclanthology.org/2022.lchange-1.9). If users want to use ipa2tokens, they can apply it before. Otherwise, we assume that the iterable that is passed is comparable.

LinguList (changes requested):
This would be the last thing I recommend, then we can merge. Thanks for your initiative.

src/lingpy/evaluate/alr.py
PatrickElmer: `tokens` and `classes` have now been remove from the `mean_edit_distance` function.

LinguList (approved):

LinguList:
I'll merge this now. The discussion has shown, that the evaluation function using edit distance should be discarded, as we'd by now do it much differently, not in the same wordlist.

LinguList:
I'll probably do so via another issue.

PR #447, opened 2023-06-07
Fixes #443: ipa2tokens raises error when being passed string with spaces

### LINKED ISSUE #443: ipa2tokens should thrown an error when being past something else but a string
This should be fixed, as we currently silently allow to handle lists, which does not make sense.

PatrickElmer:
This pull request offers a possible fix for #443. It achieves three main objectives.

**1. Removing word-initial and word-final # is removed from the `ipa2tokens` function.**

**2. Raising a value error when passing strings with spaces.**

Currently the `_list2msa` function stores `["seqs"]` as a space-separated string. This space-separated string gets passed into `ipa2tokens` in the `msa2html` function.

This pull request changes the behavior to storing 'seqs' as a list of tokens.
It also introduces a check for pre-tokenized strings in the `__init__` method of the `Multiple` class.

**3. Renaming `istring` variable to `sequence`.**

### DIFF
--- a/src/lingpy/convert/html.py
+++ b/src/lingpy/convert/html.py
@@ -425,10 +425,13 @@ def msa2html(
         for i in msa['local']:
             local[i] = '*'
 
+    tokens = msa['seqs']
+    msa['seqs'] = [' '.join(seq) for seq in msa['seqs']]
+
     # get two sorting schemas for the sequences
     if keywords['class_sort']:
 
-        classes = [tokens2class(ipa2tokens(seq), rcParams['asjp']) for seq in msa['seqs']]
+        classes = [tokens2class(seq, rcParams['asjp']) for seq in tokens]
         seqs = dict(
             [(a[1], b) for a, b in zip(
                 sorted(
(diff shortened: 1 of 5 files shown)


src/lingpy/convert/html.py
```diff
@@ -425,10 +425,13 @@ def msa2html(
         for i in msa['local']:
             local[i] = '*'
 
+    tokens = msa['seqs']
+    msa['seqs'] = [' '.join(seq) for seq in msa['seqs']]
```
PatrickElmer, line 429: Here `msa['seqs']` is reverted back to what it looked like before the PR as code further down relies on it being space-separated strings.

tests/sequence/test_sound_classes.py
```diff
@@ -17,12 +17,6 @@ def test_ipa2tokens(test_data):
     seq = 'ʰto͡i'
     assert len(ipa2tokens(seq)) == 2
 
-    seq = 'th o x t a'
```
LinguList, line 20: In this case, we may want to assert a value error, so one could use `assert raises(ValueError)`, explicit error handling, given that this is should be the new behavior, right?

LinguList (commented):
Seems fine, thanks, only the check for an error in case of space in string would be useful now.

PR #449, opened 2023-07-14
Compatibility with igraph 0.10.5 (and presumably 0.11)

xrotwang:
Supersedes #448

### DIFF
--- a/src/lingpy/algorithm/extra.py
+++ b/src/lingpy/algorithm/extra.py
@@ -201,13 +201,13 @@ def infomap_clustering(threshold, matrix, taxa=False, revert=False):
     G = igraph.Graph()
     # add vertices to the graph
     for i in range(len(matrix)):
-        G.add_vertex(i)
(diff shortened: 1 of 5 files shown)


xrotwang:
I kept this PR minimal (and somewhat hacky), because I think a more complete analysis of the issue should only be done once we modernized the code base (dropped py3.7 compat, updated packaging, ...)

LinguList:
I agree. This is perfect for now.

LinguList (approved):

LinguList:
In fact, the use of integers as ID here was in part forced or forced by my false thinking of the way node identifiers are handled in igraph. This was never really clear to me, but if you load a graph in igraph, it has (had) the numerical ID per node, while networkx always uses node label (also when reading from gml). It looked convenient in the settings used here, in the past, but in the retrospective, integration of igraph was always a bit hacky.

LinguList:
Many thanks, @xrotwang !

PR #452, opened 2023-07-21
deps: change python-igraph to igraph

szhorvat:
Please see https://github.com/igraph/python-igraph/issues/699 for why this is necessary.


LinguList:
Very cool, I did not notice this!

PR #454, opened 2023-10-13
Ipa2tk

### LINKED ISSUE #453: lxml dependency breaks pip install in Python 3.8
### Description

As the title suggests, it is currently not possible to install LingPy with ```pip install lingpy``` in Python 3.8 (and maybe older versions which I haven't tested), which fails with the following error message:
```
ERROR: Could not build wheels for lxml, which is required to install pyproject.toml-based projects.
```

This only seems to be the case for the latest version of lxml (4.9). A workaround for the problem is to install version 4.8 of lxml __before__ installing LingPy; afterwards, the installation of LingPy works as intended.

This problem does not occur in newer versions of Python; installation via pip (with dependencies to the latest version of lxml) works fine in Python 3.9 and 3.12.

### Reproduction steps

1. Create new virtual environment with Python 3.8 as base interpreter
2. Make sure no version of lxml is already installed or inherited from the base interpreter
3. Run ```pip install lingpy```

LinguList:
Fixes #453 but also fixes another problem introduced with the strict version of ipa2tokens that only accepts strings now as input (fix in pairwise module, tests added)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,7 @@ setup(
     long_description_content_type="text/markdown",
     author="Johann-Mattis List and Robert Forkel",
     author_email="info@lingpy.org",
-    version="2.6.10.dev0",
+    version="2.6.10",
     packages=find_packages(
         where="src",
         exclude=[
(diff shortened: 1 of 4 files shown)


LinguList:
This is needed in order to update lingrex to integrate fuzzy reconstructions.

LinguList:
I would merge this now and prepare a release.

PR #462, opened 2026-04-14
Update sampa.py

SimonGreenhill:
sampa.py doesn't close a file handler. 

This causes a downstream package I'm building to crash when running tests (I think the test runner garbage collects but can't find the handle) although this is claude code's interpretation so YMMV.

Here's the crash that kills pytest:
```
object address  : 0x15b1f86a0
object refcount : 3
object type     : 0x10094dbd0
object type name: ValueError
object repr     : ValueError('I/O operation on closed file.')
lost sys.stderr
```

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -15,9 +15,9 @@ jobs:
         python-version: [3.9, "3.10", 3.11, 3.12, 3.13]
 
     steps:
-    - uses: actions/checkout@v4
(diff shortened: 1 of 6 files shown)


xrotwang (approved):

xrotwang:
@SimonGreenhill fix released with lingpy 2.6.14

SimonGreenhill:
Awesome, thanks @xrotwang !

